PR #1, opened 2019-04-18
Fix Sauce on Travis and delete unused .travis.yml

TimvdLippe:
This is not working yet. It requires the new Sauce credentials in the `.travis.yml`.

Also, we are no longer testing template on specific older browsers (like Edge 13). Not sure if that is an issue.


TimvdLippe:
Merging, since it now runs on Sauce. I will fix the test failures in a separate PR.

PR #6, opened 2019-06-05
Port webcomponents/template#52 to monorepo

dfreedm:


TimvdLippe:
Never mind, we already minified that package: https://github.com/webcomponents/polyfills/commit/d145e382508457cb5417836b0d3e1ee33c1b3b83

nicksrandall:
I hope that this will include the minified version in the bundle.

PR #7, opened 2019-06-05
Port URL PRs to monorepo

azakus:
Port of webcomponents/URL#23 and webcomponents/URL#26


packages/url/urltestharness.html
```diff
@@ -47,6 +47,8 @@
       var url = bURL(expected.input, expected.base)
 
       assert_equals(url.protocol, expected.protocol, "scheme")
+      assert_equals(url.username, expected.username, "username")
+      assert_equals(url.password, expected.password, "password")
```
TimvdLippe: This test does not run on CI. Could you confirm these manual tests are now passing as well?

i-tengfei:
Any update?

PR #131, opened 2019-06-12
Stylesubtree slots

### LINKED ISSUE #11: [ShadyCSS] styleSubtree() does not consider slot elements (in polyfilled browsers)
When I try to call styleSubtree() in IE11 any element in a `<slot>` will not be styled.

This example code will not work in IE11 and older Edge Browser (Windows 10 LTSB). The Toolbar color will remain the same.

```html
  <paper-drawer-panel>
    <paper-toolbar slot="main"></paper-toolbar>
  </paper-drawer-panel>

  <button onclick="Polymer.updateStyles({ '--paper-toolbar-background': 'pink'})">Update styles</button>
```

azakus:
Remake of https://github.com/webcomponents/shadycss/pull/224

Fixes #11


azakus:
Dropped for #186, which should have identical behavior to Polymer 1.

PR #137, opened 2019-06-12
Patch `Element.outerHTML`

### LINKED ISSUE #106: Custom element not "recognized" when added with Element.outerHTML
This is happening with Firefox (not tested with others browsers that need this polyfill and of course in Chrome this works well) and with the 1.0.0-rc.1 version.
So if you add a custom element with Element.outerHTML, the custom element seems to not be "recognized" (the constructor is not called).

I wrote a simple test page where you can see the bug: http://vrac.pwet.fr/customelement_outerhtml.html

dfreedm:
Port of https://github.com/webcomponents/custom-elements/pull/149
Fixes #106


azakus:
@TimvdLippe had a non-corp email for one commit or so

PR #139, opened 2019-06-12
[Custom Elements] Throw error when `createElement` does not produce an HTMLElement

### LINKED ISSUE #112: [Custom Elements] Upgrading halts if a CE constructor throws
Upgrading custom elements halts if one element throws in its constructor.

This is because of the try/catch/re-throw here: https://github.com/webcomponents/custom-elements/blob/master/src/CustomElementInternals.js#L285

In native, element upgrading continues, with the broken element just left un-upgraded.  It would be good to match this behavior.

**Repro:**
https://jsbin.com/jopadev/edit?html,console

**Expected:** console logs
```
x-good
Error: from x-throws
x-good
```

**Actual:**
```
x-good
Error: from x-throws
```

Note that in the repro, need to wait until after `DOMContentLoaded` since the polyfill will do one add'l pass at DCL timing which would "pick up" from an error thrown before DCL.

### LINKED ISSUE #118: [Custom Elements] Error not thrown when createElement does not produce an element
When define is called with a constructor that does not extend HTMLElement, calling `document.createElement` for that name should fail.

azakus:
Port of https://github.com/webcomponents/custom-elements/pull/213

Closes https://github.com/webcomponents/custom-elements/pull/213

Fixes #112, Fixes #118

- [x] Merge https://github.com/webcomponents/polyfills/pull/149 to master and update this branch to add the Safari GC workaround to new tests.


packages/custom-elements/src/Patch/HTMLElement.js
```diff
@@ -46,7 +46,11 @@ export default function(internals) {
       const lastIndex = constructionStack.length - 1;
       const element = constructionStack[lastIndex];
       if (element === AlreadyConstructedMarker) {
-        throw new Error('The HTMLElement constructor was either called reentrantly for this constructor or called multiple times.');
+        const localName = definition.localName;
+        throw new Error('Failed to construct \'' + localName + '\': ' +
+            'This element was already constructed. The constructor may be ' +
+            'non-conformantly reentrant or might have called into the ' +
```
sorvell: The phrase  "non-conformantly reentrant" seems a bit arcane; maybe just omit that?

packages/custom-elements/tests/html/HTMLElement/constructor.html
```diff
@@ -72,25 +62,235 @@
...
+          document.createElement(localName);
+        });
+      }, (e) => {
+        e.preventDefault();
+        e.stopImmediatePropagation();
+        errors.push(e);
+      });
+      assert.equal(errors.length, 1);
+
+      assert.doesNotThrow(function() {
```
sorvell: Probably should check that there's not a special error thrown here, right?

packages/custom-elements/tests/html/HTMLElement/constructor.html
```diff
@@ -72,25 +62,235 @@
...
+        e.preventDefault();
+        e.stopImmediatePropagation();
+        errors.push(e);
+      });
+      assert.equal(errors.length, 1);
+
+      document.body.removeChild(element);
+    });
+
+  test('Constructor errors do not halt upgrades in a tree.', function() {
```
sorvell: Let's add a test that shows errors in callbacks do not halt upgrades.

bicknellr: Ok, added.

packages/custom-elements/src/CustomElementInternals.js
```diff
@@ -161,7 +164,11 @@ export default class CustomElementInternals {
       if (element.__CE_state === CEState.custom) {
         this.connectedCallback(element);
       } else {
-        this.upgradeElement(element);
+        try {
+          this.upgradeElement(element);
```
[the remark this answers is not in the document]
bicknellr: Ok, I added a separate function called `upgradeReaction` to handle this. `upgradeElement` (which I renamed to `upgradeAnElement`) is meant to match up with ['upgrade an element'](https://html.spec.whatwg.org/multipage/custom-elements.html#concept-upgrade-an-element), which doesn't report its own errors.

PR #153, opened 2019-06-18
Work around Safari GC collecting the `customElements` global.

bicknellr:
This PR works around a bug in some older versions of Safari where non-native properties on native globals set early after loading are garbage collected. The custom elements polyfill now sets a function to `window.__CE_installPolyfill` when it sees `__CE_installPolyfill` in the search string, which can be used to install the polyfill again. All the tests now check for whether or not the polyfill was garbage collected and reinstall it if was.


packages/custom-elements/tests/index.html
```diff
@@ -10,33 +10,28 @@
...
-    'html/shim.html',
-    'html/template-polyfill.html',
+    'html/reactions.html?__CE_installPolyfill=1',
+    'html/registry.html?__CE_installPolyfill=1',
+    'html/instanceof.html?__CE_installPolyfill=1',
+    'html/typescript.html?__CE_installPolyfill=1',
+    'html/babel.html?__CE_installPolyfill=1',
+    'html/closure.html?__CE_installPolyfill=1',
+    'html/upgrade.html?__CE_installPolyfill=1',
+    'html/shadow-dom.html?__CE_installPolyfill=1',
```
bicknellr: I added a wrapper page to these tests so that they could be run as modules that import the workaround function.

packages/custom-elements/src/custom-elements.js
```diff
@@ -50,4 +46,23 @@ if (!priorCustomElements ||
...
+// This is NOT public API and is only meant to work around a GC bug in older
+// versions of Safari that randomly removes the polyfill during tests. Adding
+// "__CE_installPolyfill" to the search portion of the URL will cause the
+// polyfill installation function to be added to the global object. Ideally,
+// this would use URLSearchParams but IE11 does not support it.
+//
+// This behavior can't be triggered by a scoped flag on `customElements` (like
+// `forcePolyfill`) because the flag may be garbage collected by the same bug.
+if (location.search.indexOf('__CE_installPolyfill') !== -1) {
+  window['__CE_installPolyfill'] = installPolyfill;
```
sorvell: I think this should probably just be unconditional. The code is currently making a global assumption about the search param which is basically the same as just making the global.

packages/custom-elements/tests/html/Element/setAttribute.html
```diff
@@ -43,6 +46,7 @@
 
 suite('Setting new values.', function() {
   let localName1;
+  let localName2;
```
sorvell: Was this just left out before?

packages/custom-elements/tests/html/reactions.html
```diff
@@ -1,3 +1,19 @@
+<!doctype html>
+<html>
+<head>
+<title>Custom Element Reactions</title>
+<script>
+  (window.customElements = window.customElements || {}).forcePolyfill = true;
+</script>
+<script src="../../node_modules/es6-promise/dist/es6-promise.auto.min.js"></script>
```
sorvell: Why are all these scripts new in a lot of these files?

packages/custom-elements/tests/safari-gc-bug-workaround.js
```diff
@@ -0,0 +1,27 @@
...
+// are not considered reachable and might be garbage collected. This happens
+// randomly and appears to affect not only the `customElements` global but also
+// the wrapper functions it attaches to other globals, like `Node.prototype`,
+// effectively removing the polyfill. To work around this, this function checks
+// if the polyfill is missing before running the tests. If so, it uses a
+// special global function added by the polyfill during tests (which doesn't
+// get collected) to reinstall it.
+//
+// https://bugs.webkit.org/show_bug.cgi?id=172575
+export function safariGCBugWorkaround() {
```
sorvell: Let's consider just making this the global and perhaps call it `ensureCustomElementsPolyfillInstalled`.

Users who encounter this issue would otherwise have to write this function.

sorvell: After discussion, we elected not to do this since this isn't really a feature we want to encourage using. If the native feature is available, users should be using it.

packages/custom-elements/tests/html/Element/setAttribute.html
bicknellr: Yeah, converting the test suite to a module / strict mode exposed this.

packages/custom-elements/tests/html/reactions.html
bicknellr: GitHub is showing these files as moved with additions since they're mostly the same even though they were also changed from JS to HTML. Before, they were all imported by `packages/custom-elements/tests/index.html` and run in the context of that page but now they're all separate pages.

packages/custom-elements/src/custom-elements.js
bicknellr: Updated.

PR #160, opened 2019-06-25
upgrade google-closure-compiler to 20190618.0.0

### LINKED ISSUE #148: Upgrade closure compiler

bicknellr:
Fixes #148.


bicknellr:
That sounds like good enough reason to remove it, given that a user would have to manually update their Bower dependency to point at this repo for that change to start breaking - maybe Bower has download stats somewhere? @azakus should make that call though. Also, ShadyCSS might need to be updated to build in the npm prepack script before removing the current one.

azakus:
@web-padawan that's a good point. I think it makes sense to remove it from the monorepo.

PR #165, opened 2019-06-27
Store custom element definitions at the registry, rather than the 'internals' class.

bicknellr:
Before this PR, definitions were stored on the CustomElementInternals object, which is passed to all of the patches and used to trigger reactions when those patched functions are called. Now, the definitions are lifted back into CustomElementRegistry and the CustomElementInternals object finds definitions for elements that need to be upgraded by using that element's `ownerDocument` instead. (Custom element definitions of elements that have already been upgraded are accessible via a private property, so there's no need to find a registry during the other reactions.)

This makes implementing lazy definitions simpler by allowing all of the associated behavior to stay in CustomElementRegistry. Also, moving the definitions to CustomElementRegistry is a prerequisite to prototyping scoped registries.

### DIFF
--- a/packages/custom-elements/src/CustomElementInternals.js
+++ b/packages/custom-elements/src/CustomElementInternals.js
@@ -286,7 +286,12 @@ export default class CustomElementInternals {
    */
   upgradeReaction(element) {
     try {
-      this._upgradeAnElement(element);
+      const definition = this._lookupACustomElementDefinition(
+          /** @type {!Document} */ (element.ownerDocument),
+          element.namespaceURI, element.localName);
+      if (definition) {
+        this._upgradeAnElement(element, definition);
+      }
     } catch (e) {
       this._reportTheException(e);
     }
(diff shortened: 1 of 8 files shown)


packages/custom-elements/src/CustomElementInternals.js
```diff
@@ -436,6 +392,39 @@ export default class CustomElementInternals {
...
+    // upgrading.
+    //
+    // https://html.spec.whatwg.org/multipage/custom-elements.html#look-up-a-custom-element-definition
+    //   "If document does not have a browsing context, return null."
+    //
+    // https://html.spec.whatwg.org/multipage/window-object.html#dom-document-defaultview
+    //   "The defaultView IDL attribute of the Document interface, on getting,
+    //   must return this Document's browsing context's WindowProxy object, if
+    //   this Document has an associated browsing context, or null otherwise."
+    if (!doc.defaultView && !(doc.__CE_isImportDocument && registry)) return;
```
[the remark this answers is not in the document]
bicknellr: Fixed.

packages/custom-elements/src/CustomElementInternals.js
```diff
@@ -317,7 +286,12 @@ export default class CustomElementInternals {
    */
   upgradeReaction(element) {
     try {
-      this._upgradeAnElement(element);
+      const definition = this._lookupACustomElementDefinition(
```
sorvell: Seems like this is being setup to support scoped registries. So just thinking ahead, do you think we'd need to call `getRootNode` on the element here?

bicknellr: (from offline) Exactly,`_lookupACustomElementDefinition` would likely end up using `element.getRootNode()` to get the nearest document or shadow root and use the registry there, or fall back to the registry associated with `element.getRootNode({composed: true})`.

PR #166, opened 2019-07-01
Make `reportTheException` public; record definitions pending a flush only by their local name.

bicknellr:
This PR is preparation for a lazy definition PR.

### DIFF
--- a/packages/custom-elements/src/CustomElementRegistry.js
+++ b/packages/custom-elements/src/CustomElementRegistry.js
@@ -67,9 +67,9 @@ export default class CustomElementRegistry {
 
     /**
      * @private
-     * @type {!Array<!CustomElementDefinition>}
+     * @type {!Array<string>}
      */
-    this._pendingDefinitions = [];
+    this._unflushedLocalNames = [];
 
     /**
     * @private
@@ -148,7 +148,7 @@ export default class CustomElementRegistry {
 
     this._localNameToDefinition.set(localName, definition);
     this._constructorToDefinition.set(definition.constructorFunction, definition);
-    this._pendingDefinitions.push(definition);
+    this._unflushedLocalNames.push(localName);
 
     // If we've already called the flush callback and it hasn't called back yet,
     // don't call it again.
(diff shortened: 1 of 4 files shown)


packages/custom-elements/src/CustomElementRegistry.js
```diff
@@ -67,9 +67,9 @@ export default class CustomElementRegistry {
 
     /**
      * @private
-     * @type {!Array<!CustomElementDefinition>}
+     * @type {!Array<string>}
      */
-    this._pendingDefinitions = [];
+    this._unflushedLocalNames = [];
```
sorvell: Could the `elementsWithPendingDefinitions` map and `_unflushedLocalNames` just be one thing?

packages/custom-elements/src/CustomElementRegistry.js
```diff
@@ -272,7 +271,7 @@ export default class CustomElementRegistry {
     // Resolve immediately only if the given local name has a definition *and*
     // the full document walk to upgrade elements with that local name has
     // already happened.
-    if (definition && !this._pendingDefinitions.some(d => d.localName === localName)) {
+    if (definition && this._unflushedLocalNames.indexOf(localName) === -1) {
```
sorvell: Simpler lookup if this were a Map.

bicknellr: I changed this to a Set.

bicknellr: Hmm. Let me try out the Map again though. I was trying to avoid having empty arrays sitting around before the flush but I'm not sure it's worth it after trying out Set.

bicknellr: Nevermind, I think sticking with a Set here is reasonable.

bicknellr: Welp, I flopped back...

packages/custom-elements/src/CustomElementRegistry.js
```diff
@@ -272,7 +265,7 @@ export default class CustomElementRegistry {
     // Resolve immediately only if the given local name has a definition *and*
     // the full document walk to upgrade elements with that local name has
     // already happened.
-    if (definition && !this._pendingDefinitions.some(d => d.localName === localName)) {
+    if (definition && this._elementsWithPendingDefinitions.has(localName)) {
```
bicknellr: This is wrong, I removed the `!`.

bicknellr: Fixed, also fixed the tests that were meant to catch this.

PR #171, opened 2019-07-10
Remove unused globals from externs.

aomarks:


bicknellr:
Sorry, I didn't notice this one when I did https://github.com/webcomponents/polyfills/pull/272 . They have exactly the same changes so I'll close this one.

PR #181, opened 2019-07-30
Fix issue #180: Add code to make sure the internal `inertDoc` has the…

### LINKED ISSUE #180: document.domain and <node>.innerHTML on IE11
### Description
When using the polyfills in IE11, if document.domain is set, <node>.innerHTML throws a `SCRIPT5022: WrongDocumentError` at line 84 of template/template.js.

I'm not finding any information in my various searches that shed any light on why this is happening or how to eliminate it.

Setting document.domain AFTER loading the polyfill does not change the results.

Using <node>.textContent and <node>.appendChild(document.createTextNode('New node') both work, but we are using a 3rd party web component that sets innerHTML that we may not be able to change.

This appears to be a bug in the polyfill more than an issue with the component.

#### Steps to Reproduce
Using this HTML:
```
<!DOCTYPE html>
<head>
  <title>Web component test</title>
  <script>
    document.domain = '<somedomain>.com';
  </script>
  <script src="https://unpkg.com/@webcomponents/webcomponentsjs/webcomponents-bundle.js"></script>
</head>
<body>
  Web components test.
  <div id="wc1">Initial content.</div>
  <button onclick="document.getElementById('wc1').innerHTML = 'Updated content ' + Date.now();">Update</button>
  <button onclick="addNode('div');">Add node</button>
  <script>
    function addNode(nodeName) {
      var node = document.createElement(nodeName);
      node.innerHTML = 'New node';
      document.body.appendChild(node);
    }
  </script>
</body>
```
1. Create an HTML file on your local web server using the above HTML.
2. Set `document.domain` to your domain.
3. Load the file from your web server into IE11.
4. Open the developer tools (press F12).
5. Click the 'Add node' button.

#### Expected Results
A new node at the end of the document with the text 'New node'

#### Actual Results
In the Javascript console:
SCRIPT5022: WrongDocumentError
template.js (84,11)

### Browsers Affected

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari
- [x] IE 11

kurtinatlanta:
… same domain as the real `document` object to avoid `WrongDocumentError` when setting `innerHTML` in IE11.

### Reference Issue
Fixes #180

### DIFF
--- a/packages/shadydom/src/patches/ElementOrShadowRoot.js
+++ b/packages/shadydom/src/patches/ElementOrShadowRoot.js
@@ -15,6 +15,11 @@ import {clearNode} from './Node.js';
 /** @type {!Document} */
 const inertDoc = document.implementation.createHTMLDocument('inert');
 
+// Make sure that the domains between the inertDoc and the 'real' document object match in order
+if (inertDoc.domain !== document.domain) {
+  inertDoc.domain = document.domain;
+}
+
 export const ElementOrShadowRootPatches = utils.getOwnPropertyDescriptors({
 
   /** @this {Element} */
--- a/packages/shadydom/src/patches/ElementOrShadowRoot.js
+++ b/packages/shadydom/src/patches/ElementOrShadowRoot.js
@@ -15,11 +15,6 @@ import {clearNode} from './Node.js';
 /** @type {!Document} */
 const inertDoc = document.implementation.createHTMLDocument('inert');
 
-// Make sure that the domains between the inertDoc and the 'real' document object match in order
-if (inertDoc.domain !== document.domain) {
-  inertDoc.domain = document.domain;
-}
-
 export const ElementOrShadowRootPatches = utils.getOwnPropertyDescriptors({
 
   /** @this {Element} */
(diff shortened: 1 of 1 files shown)


kurtinatlanta:
@googlebot I signed it!

jason0x43:
The current fix works, but only when `document.domain` is defined before the webcomponentsjs bundle is loaded. Possibly a `inertDoc`'s domain should be updated in `set innerHTML` instead (or in addition)?

kurtinatlanta:
I wasn't sure whether the check was needed with each call, but that sounds a bit more resilient. I'll give this a try.

janabimustafa:
Hello! @jason0x43 @kurtinatlanta I am running into the same issue as #180. When will this PR get merged?

cekvenich:
And released please, eg https://cdn.jsdelivr.net/npm/@webcomponents/webcomponentsjs@2.4.0/webcomponents-loader.js

So that we can get it the way we used to before.

znewton:
This would be super helpful! We have a lot of pages setting `document.domain` to support iFrames, but we are also trying to add support for the webcomponent implementation of our company's design system.

Is there anything needed to help push this along?

justinfagnani:
@kurtinatlanta I'm a bit unfamiliar with the area of the code, but at it'd be nice to have a test that fails currently on IE11 and passes after this change.

packages/shadydom/src/patches/ElementOrShadowRoot.js
```diff
@@ -37,6 +37,16 @@ export const ElementOrShadowRootPatches = utils.getOwnPropertyDescriptors({
       this[utils.NATIVE_PREFIX + 'innerHTML'] = value;
     } else {
       clearNode(this);
+      // Make sure that the domain attributes of inertDoc and the document for
+      // are the same in order to avoid WrongDocumentError in IE11.
+      if ((inertDoc.domain.indexOf(document.domain) >= 0) && (inertDoc.domain !== document.domain)) {
```
sorvell: Doing this check on every call to `innerHTML` is likely bad for performance. It would be better to override the setter for `domain` on document and sync the inertDoc domain there.

piyushsh:
Any idea when this fix is going to be in? This is something we are also struggling with.

Issue #8, opened 2019-06-06
ScopingShim and nested calc issues in IE and Edge <=15


wessberg:
Hi there. Thanks for writing ShadyCSS and for bringing the web forward.

# Problem 
Unfortunately, in IE and Edge <= 15, nested `calc`s (i.e. `calc` inside of another `calc`) are not supported. When using and especially combining CSS custom properties, it is easy to end up combining multiple CSS Custom Properties in `calc` expressions. Some of those may include `calc`s on their own, which leads to this problem.

Developers can get around this by either rewriting their CSS to avoid nested `calc`s, (which can be difficult when using CSS Custom Properties to perform dynamic styling) or use tooling to get around it. For example, in the [postcss](https://github.com/postcss/postcss) ecosystem, combining the [`postcss-custom-properties`](https://github.com/postcss/postcss-custom-properties) and [`postcss-calc`](https://github.com/postcss/postcss-calc) plugins helps reducing and avoiding nested `calc`s to fix styling issues in these browsers.

*However*, not everything can be fixed at compile-time. For example, consider the following:

```css
/* styling applied to 'my-element' */
:host {
    top: calc(var(--foo) - var(--bar));
}

/* somewhere else in some CSS, those properties are set: */
my-element {
    --foo: calc(1px * 2);
    --bar: calc(2px * 2);
}
```

In this example, both `--foo` and `--bar` are dynamic properties, and each of them include a `calc`.

After applying the ScopingShim, the declaration in the CSSOM will be:
```css
top: calc(calc(1px * 2) - calc(2px * 2));
```

Which simply won't work in those browsers.

# Proposed Solution

There's been some effort going into fixing this issue at runtime such as [`css-vars-ponyfill`](https://www.npmjs.com/package/css-vars-ponyfill) that fixes nested `calc`s by default, and I think it would be very valuable for ShadyCSS to do the same thing. I expect some performance overhead since parsing and reducing `calc` rules isn't trivial, but I think it is well worth it.

I've submitted a PR: https://github.com/webcomponents/shadycss/pull/242

wessberg:
I would appreciate if this issue is kept open, as there is a PR ready and this causes issues in IE and Edge <= 15. By merging in the open PR, better interoperability will be achieved with these older browsers.

wessberg:
[...]

PR #196, opened 2019-09-09
[ShadyCSS] Parse and flatten nested calc()

wessberg:
### Reference Issue
https://github.com/webcomponents/polyfills/issues/8

Migrated from: https://github.com/webcomponents/shadycss/pull/242

#### Original description:

This PR adds a minimal CSS calc() parser strictly for flattening nested calc() expressions which may be the result of shimming CSS Custom Properties at runtime. This is important because browsers such as IE and Edge <= 15 doesn't support nested calc() which leads to broken styling under some circumstances as is outlined in #241.

The implementation use a heuristic to take a fast path for text that include less than two calc()s to avoid unnecessary work.

I've added Closure Compiler comments and there are no warnings.

### DIFF
--- a/packages/shadycss/src/calc-parse.js
+++ b/packages/shadycss/src/calc-parse.js
@@ -1,0 +1,227 @@
+/**
+ @license
+ Copyright (c) 2019 The Polymer Project Authors. All rights reserved.
+ This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+ The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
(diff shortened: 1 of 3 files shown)


wessberg:
As I wrote last time this was marked as stale:

> I would appreciate if this issue is kept open, as there is a PR ready and this causes issues in IE and Edge <= 15. By merging in the open PR, better interoperability will be achieved with these older browsers.

At this point, this issue and PR has been open for _years_ with no response from the maintainers. I've had the web components polyfills served based on this PR for years via polyfill.app, which improves compatibility with these browsers visually

PR #199, opened 2019-09-10
Fix issue with custom-style ordering under Custom Property shim

azakus:
When using shim'd Custom Properties, the ordering of custom styles must
be in document order, matching the CSS cascade order when using native
Custom Properties.

The order of custom-styles internally was only in "upgrade" order, which
could be out of date if a custom-style is created after loading.

### DIFF
--- a/packages/shadycss/src/scoping-shim.js
+++ b/packages/shadycss/src/scoping-shim.js
@@ -22,7 +22,7 @@ import {flush as watcherFlush, getOwnerScope, getCurrentScope} from './document-
 import templateMap from './template-map.js';
 import * as ApplyShimUtils from './apply-shim-utils.js';
 import {updateNativeProperties, detectMixin} from './common-utils.js';
-import {CustomStyleInterfaceInterface} from './custom-style-interface.js'; // eslint-disable-line no-unused-vars
+import {CustomStyleInterfaceInterface, CustomStyleProvider} from './custom-style-interface.js'; // eslint-disable-line no-unused-vars
 
 /** @type {!Object<string, string>} */
 const adoptedCssTextMap = {};
@@ -229,6 +229,7 @@ export default class ScopingShim {
       return;
     }
     if (!nativeCssVariables) {
+      this._validateCustomStyleOrder(customStyles);
       this._updateProperties(this._documentOwner, this._documentOwnerStyleInfo);
       this._applyCustomStyles(customStyles);
       if (this._elementsHaveApplied) {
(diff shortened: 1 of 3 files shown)


packages/shadycss/src/scoping-shim.js
```diff
@@ -240,6 +241,29 @@ export default class ScopingShim {
...
+  /**
+   * Validate ordering of custom styles for Custom Property shim
+   * @param {!Array<!CustomStyleProvider>} customStyles
+   */
+  _validateCustomStyleOrder(customStyles) {
+    const styles = customStyles.map(c => this._customStyleInterface['getStyleForCustomStyle'](c));
+    // sort styles in document order
+    styles.sort((a, b) => {
+      // use `b.compare(a)` to be more straightforward
+      const position = b.compareDocumentPosition(a);
```
bicknellr: Are custom styles moved to the main document when using real HTML Imports? IIRC, `compareDocumentPosition` needs the nodes being compared to have the same shallow root node to return a useful value.

packages/shadycss/src/scoping-shim.js
```diff
@@ -240,6 +241,29 @@ export default class ScopingShim {
     }
     this._customStyleInterface['enqueued'] = false;
   }
+  /**
+   * Validate ordering of custom styles for Custom Property shim
+   * @param {!Array<!CustomStyleProvider>} customStyles
+   */
+  _validateCustomStyleOrder(customStyles) {
```
bicknellr: This function seems to be establishing the order itself or repairing it rather than validating - maybe a different name?

azakus:
Chrome failing in tests because travis is having a hard time installing the latest version, and the "default" version is too old for WCT. 😡

packages/shadycss/src/scoping-shim.js
bicknellr: (from offline) Yes, custom styles move themselves into the main document: https://github.com/Polymer/polymer/blob/master/lib/elements/custom-style.js#L93-L105

packages/shadycss/src/scoping-shim.js
azakus: Yeah, I'll just rename it to `_reorderCustomStylesRules`

azakus: Done. PTAL

packages/shadycss/tests/custom-style-ordering.html
```diff
@@ -0,0 +1,81 @@
...
+    test('custom styles apply properties in document order', function() {
+      const el = document.createElement('x-foo');
+      document.body.appendChild(el);
+      if (window.ShadyDOM) {
+        ShadyDOM.flush();
+      }
+      assertComputed(el, 'rgb(255, 0, 0)', 'background-color');
+      const tmpl = document.querySelector('template#late-style');
+      // append style to head, before custom-style in body
+      document.head.appendChild(tmpl.content.cloneNode(true));
```
bicknellr: IMO it would be a bit clearer to explicitly insert this style relative to the other but it's not critical.

azakus: Done!

Issue #55, opened 2019-02-07
[ShadyCSS] Add standalone <custom-style> like element for scoping main document styles


justinfagnani:
We've had a request from a few people to be able to add scope-respecting main-document styles when not using Polymer, like with LitElement.

jrobinson01:
I've started to work on this but running into a snag. I've created the standalone element here: https://github.com/jrobinson01/polyfills/blob/custom-style-element/packages/shadycss/src/custom-style.js 
and then import it here:
https://github.com/jrobinson01/polyfills/blob/custom-style-element/packages/shadycss/entrypoints/custom-style-interface.js#L16
which in theory, would allow me to use `<custom-style>` and have it work as long as the custom-style-interface polyfill is loaded. I'm hitting a snag in IE11 though, which complains that `customElements is not defined`. Is there something I'm missing in the build step?

jrobinson01:
I've moved past the above issues by waiting for the customElements registry to exist, but I'm still hitting the same problem that I started with this morning, which was that using the custom-style-interface polyfill in my project, which uses the open-wc rollup config to load polyfills on-demand, the css variables are not being applied to my elements at all. I'm assuming it's some sort of race condition as it seems that after some point during the page load, calling `CustomStyleInterface.addCustomStyles({getStyle: function() { return document.querySelector('style')})` has zero affect. This is proven by running the above code in the console after the page has loaded.

justinfagnani:
cc @azakus and @sorvell who know this better.

I'm glad you're trying this out @jrobinson01. Would you be willing to contribute it to the https://github.com/webcomponents polyfills?

jrobinson01:
@justinfagnani that's the plan.

jrobinson01:
After sleeping on this, I'm not sure I like the custom-style element being in the polyfill. It's convenient, but also feels like maybe it should be in it's own package.

justinfagnani:
Oh, I think it should be its own package, but I conceptually see this as part of the polyfills collection, since it implements style scoping at the document scope level.

jrobinson01:
Yeah exactly. I'll follow the structure of the other packages and see what it looks like. Closing my PR and will create a new one.

PR #205, opened 2019-09-18
[ShadyCSS] custom-style element

jrobinson01:
### Reference Issue
#55 

Adds a standalone `<custom-style>` element that works with the custom-style-interface to enable css variables to be used at the document level.

### Caveats
I fought with this for quite a while. I'm dynamically loading polyfills at runtime, and discovered that in order for this to work, the webcomponents polyfill must load before the custom-style-interface polyfill. I don't have a great solution to that problem, but in my project I ended up just concatenating the two together and loading that. 🤷‍♂


jrobinson01:
Closing in favor of moving the custom-style element into it's own package.

PR #217, opened 2019-10-17
docs: add lerna badge

armujahid:
These polyfills are organized as a monorepo using [Lerna](https://lerna.js.org/). So I think we should either add lerna badge
[lerna](https://lerna.js.org/) or some kind of acknowledgment in Readme (If addition of badge is not acceptable).


armujahid:
Kindly review this :)

aomarks:
Thanks for the PR, but we use dozens of tools, and I don't really think it makes sense for Lerna or anything else to be advertised in our README.

PR #206, opened 2019-09-19
shady-css-scoped element package

jrobinson01:
Adds a new package to expose a <custom-styles> element that wraps the custom-styles-interface polyfill.

### Reference Issue
fixes #55

PR #220, opened 2019-10-19
Don't pre-patch CDATASection since it fails on old IE versions.

kevinpschaaf:
Fixes failure on IE11 used on internal CI.


kevinpschaaf:
CI failure is only due to addition of `shady-css-scoped` element, which was merged with failures on polyfilled browsers; see https://github.com/webcomponents/polyfills/pull/206#issuecomment-544606288.

PR #221, opened 2019-10-22
Update tests to include polyfills required. Fixes #55.

### LINKED ISSUE #55: [ShadyCSS] Add standalone <custom-style> like element for scoping main document styles

kevinpschaaf:
The original test doesn't include any polyfills, so wasn't testing the polyfilled behavior (the point of the element), and was only passing on browsers with native CE & SD.

### DIFF
--- a/packages/shady-css-scoped-element/package-lock.json
+++ b/packages/shady-css-scoped-element/package-lock.json
@@ -25,16 +25,34 @@
       "integrity": "sha512-IxzUe6YzaORzUksafHAXHprV29YncOJgr0+1zNAifl0/f+cb5iAd4IWUrnsnVFHG5UGTLjvis5RgV6vvIZPDrA==",
       "dev": true
     },
+    "@webcomponents/custom-elements": {
+      "version": "1.3.0",
+      "resolved": "https://registry.npmjs.org/@webcomponents/custom-elements/-/custom-elements-1.3.0.tgz",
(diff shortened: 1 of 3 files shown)


kevinpschaaf:
There's an open question here around who loads `shadycss/entrypoints/custom-style-interface.js`.  It would be most convenient if the element loads it, I however _believe_ that means it will be incompatible with using a legacy Polymer `<custom-style>` element on the same page, since `<custom-style>` does include it (although an app is unlikely to ever need to do this, a `<custom-style>` can be loaded transitively via legacy e.g. paper-elements).

@azakus Do you have any insight?  Did you consider how this would work?

kevinpschaaf:
One option would be to make the global `window.ShadyCSS.CustomStyleInterface` first-in wins (i.e. only do the assignment here if it doesn't yet exist: https://github.com/webcomponents/polyfills/blob/master/packages/shadycss/entrypoints/custom-style-interface.js#L82) since it's not much code.

bicknellr:
(Asking because I'd like to get the build green again before making a new release.)

Is there a reason why it's not OK for this element to import it directly, as `<custom-style>` does [here](https://github.com/Polymer/polymer/blob/dc20feca22ed6915eb67a6df2a8475ceba295de7/lib/elements/custom-style.js#L10)? Modules only run once so, AFAIK, this should be fine. Or maybe the worry here is about what happens when npm 'resolves' conflicts by referencing two different versions?

bicknellr:
BTW, CI tests for this package were disabled in https://github.com/webcomponents/polyfills/pull/226 , so I'm not going to consider this blocking a release anymore.

jrobinson01:
Where did we end up with getting this released?

PR #225, opened 2019-11-01
[ShadyDOM] When checking if a node is a shadyRoot, avoid looking at e…

### LINKED ISSUE #215: IE11 performance question
### Description

I've read that ["The polyfill gets slower as the size of your page and number of custom element definitions increases."](https://github.com/webcomponents/polyfills/tree/master/packages/custom-elements#customelementspolyfillwrapflushcallback) I wrote a simple test for IE11 to understand this better and have questions (below) about whether anything can be done to improve performance.

#### Live Demo

https://jsbin.com/nebitij/3/edit?html,console,output

```html
<!DOCTYPE html>
<html>
  <head>
    <!-- uncomment/comment this: -->
    <script src="https://unpkg.com/@webcomponents/webcomponentsjs@2.3.0/webcomponents-loader.js"></script>
  </head>
  <body>
    <label for="nodes"># of nodes to add:</label>
    <input type="text" name="nodes" id="nodes" value="10000" />
    <button onClick="addNodes()">Add Nodes</button>

    <div id="output"></div>

    <script>
      function addNodes() {
        var start = performance.now();

        var output = document.getElementById("output");

        var numberOfNodes = Number(document.getElementById("nodes").value);

        for (var i = 0; i < numberOfNodes; i++) {
          var div = document.createElement("div");
          div.innerHTML = "test";

          output.appendChild(div);
        }

        var end = performance.now();
        var timeInMilliseconds = end - start;
        var seconds = timeInMilliseconds / 1000;

        alert("Nodes added: " + seconds.toFixed(2) + " seconds");
      }
    </script>
  </body>
</html>

```
#### Steps to Reproduce

1. Run the page above in IE11
2. Click "Add Nodes." 10,000 test divs will get added and an alert will popup showing how many seconds it takes to complete
3. Comment out the polyfill loader and reload page
4. Click "Add Nodes" again and wait for alert message.
5. Observe the time difference with polyfill and without

#### Actual Results

Adding the 10,000 test divs takes less than 1 second without the polyfill for me on my macbook pro (running IE11 on virtualbox). It takes around 6 seconds with the polyfill. As I add more nodes, clicking the button and adding the same # of additional nodes takes longer and longer as the total # of page nodes grows.

#### Expected Results

Even without custom elements in my test, I expected all the wrapped/patched dom api calls to add some overhead to all dom calls and be slow in IE11, but I have questions:

#### Questions

1) Which part of the polyfill is contributing most to the performance impact? Would it be significantly faster without ShadyCSS/ShadyDOM if we only wrote custom elements that didn’t use shadow dom? If so, is there an easy way to make the loader not load ShadyCSS/ShadyDOM so I still get all the feature detection logic for other polyfills? I saw there used to be a "webcomponents-lite" bundle that seemed to not have shadow dom, but it's unclear if that's for v0 or v1

2) Is there any way to "scope" this performance hit to just a specific part of the page that uses custom elements/shadow dom/templates/etc. and not affect other dom calls that don't need custom elements/shadow dom/etc.? I suspect not since you have to wrap/patch all these global dom apis, but I am doing my due diligence so we can determine if this is viable to use at work in a production website.

#### Background

We recently used the polyfill bundle in a production website and had some significant performance problems in IE11. The particular page was rather complex (5700 dom nodes) and started taking over 45 seconds to load on my macbook pro on IE11 (it’s a SPA page that gets built up client side after the polyfill has been loaded). Removing the polyfill made it go back to normal. 

### Browsers Affected

- [ ] Chrome
- [ ] Firefox
- [x] Edge
- [ ] Safari
- [x] IE 11

sorvell:
…xpando property on document

Fixes #215.

This is extremely slow on some versions of IE/Edge. Instead check if the node is a DocumentFragment first.

Possibly related to the getter on Document defined here: https://html.spec.whatwg.org/multipage/dom.html#document and https://html.spec.whatwg.org/multipage/dom.html#dom-document-nameditem

### DIFF
--- a/packages/shadydom/src/utils.js
+++ b/packages/shadydom/src/utils.js
@@ -35,7 +35,7 @@ export const isTrackingLogicalChildNodes = (node) => {
 }
 
 export const isShadyRoot = (obj) => {
-  return Boolean(obj instanceof DocumentFragment && obj._localName === 'ShadyRoot');
+  return obj instanceof ShadowRoot;
 }
 
 export const hasShadowRootWithSlot = (node) => {
--- a/packages/shadydom/src/utils.js
+++ b/packages/shadydom/src/utils.js
@@ -35,7 +35,7 @@ export const isTrackingLogicalChildNodes = (node) => {
 }
 
 export const isShadyRoot = (obj) => {
-  return Boolean(obj._localName === 'ShadyRoot');
+  return Boolean(obj instanceof DocumentFragment && obj._localName === 'ShadyRoot');
 }
 
 export const hasShadowRootWithSlot = (node) => {
--- a/packages/shadydom/src/utils.js
+++ b/packages/shadydom/src/utils.js
@@ -34,9 +34,7 @@ export const isTrackingLogicalChildNodes = (node) => {
   return (nodeData && nodeData.firstChild !== undefined);
 }
 
-export const isShadyRoot = (obj) => {
-  return obj instanceof ShadowRoot;
-}
+export const isShadyRoot = obj => obj instanceof ShadowRoot;
 
 export const hasShadowRootWithSlot = (node) => {
   const nodeData = shadyDataForNode(node);
--- a/packages/shadydom/src/attach-shadow.js
+++ b/packages/shadydom/src/attach-shadow.js
@@ -60,11 +60,6 @@ class ShadyRoot {
   }
 
   _init(host, options) {
-    // NOTE: set a fake local name so this element can be
-    // distinguished from a DocumentFragment when patching.
-    // FF doesn't allow this to be `localName`
-    /** @type {string} */
-    this._localName = SHADYROOT_NAME;
     // root <=> host
     this.host = host;
     /** @type {!string|undefined} */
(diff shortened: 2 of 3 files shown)


packages/shadydom/src/utils.js
```diff
@@ -35,7 +35,7 @@ export const isTrackingLogicalChildNodes = (node) => {
 }
 
 export const isShadyRoot = (obj) => {
-  return Boolean(obj._localName === 'ShadyRoot');
+  return Boolean(obj instanceof DocumentFragment && obj._localName === 'ShadyRoot');
```
kevinpschaaf: Can we use `instanceof ShadowRoot` and eliminate the `_localName` check altogether?

sorvell: Yeah, I believe that should work now. Will try and see if that's green in testing.

sorvell:
Tested Polymer to be passing with this change.

mkay581:
Thanks so much for doing this @sorvell. 🤗  Is there an ETA on when this would make it into the next `@webcomponents/shadydom` release?

web-padawan:
@sorvell @kevinpschaaf we would be also happy to test a new `@webcomponents/webcomponentsjs` release that would include this fix. Would you have time to release this fix?

It's important for us as we'd like to see how much it helps `vaadin-grid` in Edge. Assuming the performance regression with `insertBefore` is fixed now, that could help us a lot.

PR #231, opened 2019-11-09
Update README.md: Removed text that says ShadowDOM is required; I’m assuming that is a typo.

kito99:
Removed text that says ShadowDOM is required; I’m assuming that is a typo.


aomarks:
I think it's technically correct, because the DOM half of shadow support is provided by the shadydom package, whereas shadycss implements the CSS scoping half.

PR #234, opened 2019-11-19
Remove references to `customElements.nativeHTMLElement`.

bicknellr:
I think this is left over from an older version of the custom elements polyfill. These are the only places in the repo that reference anything called `NativeHTMLElement` or `nativeHTMLElement`.


bicknellr:
Yeah, this should still go in IMO.

bicknellr:
Looks like Sauce is struggling to connect / start jobs right now. I'll keep restarting it until it hits a legit pass / fail from the tests.

Issue #105, opened 2019-06-07
`instanceof Function` doesn't work for functions from other realms.


klarkc:
Hello, I am having some troubles with jsdom. I found this bug tmpvar/jsdom#1737 that is related to this piece of code: https://github.com/webcomponents/custom-elements/blob/master/src/CustomElementRegistry.js#L64

It seems that lot of _sandboxing_ by jsdom is necessary to emulate the instanceof behavior.

This check is related to the specs? Every custom element class should really be instanceof Function?

phactum-mnestler:
I'm now facing this issue while developing a Firefox extension - within a content script, the `instanceof Object` check fails for my custom elements. The error reported is `TypeError: The custom element constructor's prototype is not an object.` although the same CE works outside of an extension script.

(As for why I don't use the native Firefox `customElements`: there are a few inconsistencies in Firefox with regards to Vue.js custom elements in content scripts, so I had hoped the polyfill would smooth these over.)

PR #249, opened 2020-01-20
CERegistry.define use typeof instead instanceof

cronon:
element definition may come from another realm or Function or Object
globals may be sandboxed like in jsdom

original PR
https://github.com/webcomponents/custom-elements/pull/180

fixes
https://github.com/webcomponents/polyfills/issues/105


cronon:
@stale, well the problem still exists here. Although if I'm not mistaken couple of months ago jsdom started to support Custom Elements by themselves

PR #256, opened 2020-01-31
Replace polyfills Travis with GitHub Actions

aomarks:
Finishes in about ~20 min vs ~30 min with Travis. I broke it into 3 jobs: lint, wct-local, and wct-sauce. The jobs run in parallel and make it easier/faster to see which steps passed/failed.


.github/workflows/tests.yml
```diff
@@ -0,0 +1,71 @@
...
+    - name: NPM install
+      run: npm ci
+
+    - name: Lerna bootstrap
+      run: npm run bootstrap
+
+    - name: Build
+      run: npm run build
+
+    - name: WCT sauce (Safar/Edge/IE11/Chrome 41)
```
dfreedm: Safari*

PR #263, opened 2020-02-05
Gather event properties from the correct prototypes.

### LINKED ISSUE #261: [ShadyDOM]: Event attributes from Document incorrectly copied to HTMLElement
In order to ensure attributes like `onclick` work correctly with the polyfill, the corresponding properties are patched with ShadyDOM specific behavior. 

However, the current code writes all `on*` properties to `HTMLElement.prototype` from `Document.prototype` instead of from `HTMLElement.prototype` as it should. This results in some extra `on*` attributes being incorrectly copied like `onreadystatechange`.

bicknellr:
Fixes #261

### DIFF
--- a/packages/shadydom/tests/shady.html
+++ b/packages/shadydom/tests/shady.html
@@ -11,6 +11,16 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 <html>
 <head>
   <meta charset="utf-8">
+  <script>
+    window.NATIVE = {
+      ElementEventPropertyNames:
+          Object.getOwnPropertyNames(Element.prototype)
+          .filter(x => x.substring(0, 2) === "on"),
+      HTMLElementEventPropertyNames:
+          Object.getOwnPropertyNames(HTMLElement.prototype)
+          .filter(x => x.substring(0, 2) === "on"),
+    };
+  </script>
   <script src="wct-browser-config.js"></script>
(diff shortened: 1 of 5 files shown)


packages/shadydom/tests/shady.html
```diff
@@ -721,6 +731,24 @@
   });
 })
 
+suite('Event property patches should not add new properties', () => {
```
sorvell: It doesn't look like we have tests for "on*" attributes working. Would you mind adding one? We want to ensure that it's running through the polyfill and to do so you should be able to look at `composedPath`.

bicknellr: There's a separate file called `event-path.html` that seems to test some of the on* properties:

phases:
https://github.com/webcomponents/polyfills/blob/1aa89fd85ef7f24660233402919a510a1e83aa22/packages/shadydom/tests/event-path.html#L569-L585

targets:
https://github.com/webcomponents/polyfills/blob/1aa89fd85ef7f24660233402919a510a1e83aa22/packages/shadydom/tests/event-path.html#L636-L666

I didn't put these in there because this didn't really seem related to event paths though. Are those the kinds of tests you were looking for? / Should I move these?

sorvell: Sorry, this doesn't make sense. The `on*` attributes are not supported since they are parser driven. Imo, adding support for this is not important so let's skip it.

PR #264, opened 2020-02-06
Share each browser process across all package tests

aomarks:
Previously during testing, we launched wct separately for each package, which means spinning up a new instance of each browser for each package.

Now, we use a top-level wct.conf that aggregates all package tests, which means we re-use instances across all packages.

This should make running all polyfill tests faster, especially for Sauce which was spinning up a new VM for every browser + package combination.

Just needed one change to an html-imports test, which was using an absolute URL that was assuming that the testfiles were being served from the html-imports package root.

In the most recent two runs, this cut total test time from ~18m to ~13m.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2169,6 +2169,15 @@
       "integrity": "sha512-h+hqYkL+tQV/y2ESD5gFXMl5z4cC+XY1jTlBeGSBaTcj3VbB5OBEScbvRXm63NcEbBneQQYbHfBAXAkF9i9wIA==",
       "dev": true
     },
+    "@polymer/polymer": {
+      "version": "3.3.1",
+      "resolved": "https://registry.npmjs.org/@polymer/polymer/-/polymer-3.3.1.tgz",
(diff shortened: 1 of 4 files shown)


bicknellr:
The speedup is great but I'm honestly more excited that it reduces the number of times sauce connect could fail to open a tunnel from N to 1!

bicknellr:
Most of the packages' local `wct.conf.json`s have a setting like `"environmentImports": []` (though not consistently nested at the same point?), what is that / should we should include it here?

aomarks:
> Most of the packages' local `wct.conf.json`s have a setting like `"environmentImports": []` (though not consistently nested at the same point?), what is that / should we should include it here?

Looks like it's a way to inject additional scripts: https://github.com/Polymer/web-component-tester/blob/master/browser/environment.ts#L43

It defaults to `[]` when npm is true, which it is for us: https://github.com/Polymer/tools/blob/281fc194662396470091ac0ee21d94117257f89b/packages/web-component-tester/browser/config.ts#L68

So I don't think we need it.

Issue #260, opened 2020-02-05
failing shadycss scoping-api and disable-runtime tests


aomarks:
The following tests are failing:

- shadycss/
  - tests/
    - scoping-api.html
      - mutation observer is disabled
      - currentScopeForNode
    - disable-runtime.html
      - no scoping occurs

These failures had been masked since https://github.com/webcomponents/shadycss/pull/210 (Sep 19, 2018), because since then shadycss has had a pinned semver on shadydom, which meant lerna was not creating a local link for that dependency (lerna won't create links if the semvers are not compatible).

PR #265, opened 2020-02-06
Unpin shadycss dependency on shadydom and fix issues

aomarks:
Since Sep 2018 (https://github.com/webcomponents/shadycss/pull/210), ShadyCSS has had a pinned dependency on ShadyDOM. That meant that `lerna link` did not create a symlink between ShadyCSS and ShadyDOM, so the tests in ShadyCSS that make use of ShadyDOM were not running against the latest version, but a specific older version on NPM instead.

A couple things broke and are also addressed in this PR:

- 3 ShadyCSS tests have been failing. In the interest of incremental progress, I just marked them as `skip` and filed https://github.com/webcomponents/polyfills/issues/260 to track. I haven't yet looked into what's really going on here.

- Don't rely on `Object.assign` / `Array.from` in ShadyDOM. While the main webcomponentsjs bundles include polyfills for these APIs, the isolated `shadydom.min.js` file does not, so it was not compatible with older browsers out-of-the-box. Rather than adding those polyfills, I opted to just include local versions in the utils module. The `Object.assign` usage was added in https://github.com/webcomponents/polyfills/pull/189.

### DIFF
--- a/packages/shadycss/package.json
+++ b/packages/shadycss/package.json
@@ -38,7 +38,7 @@
   "devDependencies": {
     "@webcomponents/custom-elements": "^1.3.2",
     "@webcomponents/html-imports": "^1.2.2",
-    "@webcomponents/shadydom": "1.1.2",
+    "@webcomponents/shadydom": "^1.1.2",
     "@webcomponents/template": "^1.4.1",
     "@webcomponents/webcomponents-platform": "^1.0.0",
     "es6-promise": "^4.2.5",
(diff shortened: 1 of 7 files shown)


aomarks:
> I still think we should consider it a bug in either our build process / description of our supported feature-set floor / something else that we need to do the `Object.assign` and `Array.from` workaround, since we're already authoring in something >=ES2015 given `class`, etc., but this works for now.

The JS we distribute is closure compiled to ES5, so I think syntax like class is no problem in terms of our authoring format. Normally Object.assign/Array.from would be fine too, except we explicitly tell Closure not to include polyfills for them.

PR #270, opened 2020-02-12
Fix detection of forked PRs in testing GitHub action

aomarks:
Via https://github.community/t5/GitHub-Actions/Distinguish-between-forked-PR-and-own-PR/m-p/36198/highlight/true#M2477

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -45,7 +45,7 @@ jobs:
   wct-sauce:
     # Skip if this is from a forked repository because we can't access Sauce
     # secrets.
-    if: github.repository == 'webcomponents/polyfills'
+    if: github.event.pull_request.base.repo.id == github.event.pull_request.head.repo.id
 
     runs-on: ubuntu-latest
 


bicknellr:
Wouldn't this allow people to read our sauce credentials by forking and then changing the `test-sauce` script to something like `echo $SAUCE_USERNAME $SAUCE_ACCESS_KEY`?

bicknellr:
Oh someone mentions in that thread:

> Yea, you won't be able to use secrets in Actions run from forks. So if you use secrets to publish this image, it won't succeed for forks anyway.

So I don't think I understand how GitHub actions / secrets work together.

aomarks:
> Wouldn't this allow people to read our sauce credentials by forking, changing the `test-sauce` script to something like `echo $SAUCE_USERNAME $SAUCE_ACCESS_KEY`, and submitting it as a PR?

PRs from forks can never access secrets. The environment variables are not set.

Before this PR, we always ran the sauce job, and in the case of a forked PR it would get all the way to WCT and then fail because the secret environment variables were not set.

After this PR, we skip the sauce job for forks. This is the behavior we used here on Travis, and in our other repos. The reason it wasn't already working this way was that I made a mistake in the first PR I sent that converted from Travis to GitHub actions.

aomarks:
> So I don't think I understand how GitHub actions / secrets work together.

The way secrets work is that you configure them through the GitHub repo settings page, and then they will be set as environment variables when actions run on branches/PRs originating from this repo. For forked repos, they are not set.

bicknellr:
Oh, is `github.repository == 'webcomponents/polyfills'` always true for PRs submitted here (maybe because `github.repository` refers to the base?) and the new condition changes it to run only when the PR base and head repos are the same? (i.e. `github.event.pull_request.base.repo.id` is constant for PRs submitted here, so the new condition is only true when the branch is also already in this repo?)

aomarks:
> Oh, is `github.repository == 'webcomponents/polyfills'` always true for PRs submitted here (maybe because `github.repository` refers to the base?)

yep

> and the new condition changes it to run only when the PR base and head repos are the same? (i.e. `github.event.pull_request.base.repo.id` is constant for PRs submitted here, so the new condition is only true when the branch is also already in this repo?)

yep

PR #273, opened 2020-02-19
Merge master to html-imports

bicknellr:


bicknellr:
The commits the CLA bot was worried about in this PR are from https://github.com/webcomponents/polyfills/pull/206, which was squashed and ended up with the author's other email attached to the squashed merge commit.

PR #274, opened 2020-02-20
Add license header to Externs.ts

bicknellr:


bicknellr:
Opened https://github.com/webcomponents/polyfills/issues/275 to track that question.

PR #292, opened 2020-04-13
Convert webcomponentsjs platform to TypeScript

rictic:
I made as few changes as possible, goal here is just the initial work to get things building in TS.


packages/webcomponentsjs/ts_src/platform/custom-event.ts
```diff
@@ -0,0 +1,101 @@
...
+      (<T extends unknown>(inType: string, params?: CustomEventInit<T>) => {
+        params = params || {};
+        const e =
+            /** @type {!CustomEvent} */ (document.createEvent('CustomEvent'));
+        e.initCustomEvent(
+            inType, Boolean(params.bubbles), Boolean(params.cancelable),
+            params.detail);
+        return e;
+      }) as unknown as typeof CustomEvent;
+  window['CustomEvent'].prototype = window.Event.prototype as CustomEvent;
```
bicknellr, line 76: What's the reasoning behind which references use `obj.x` vs `obj['x']`? `Event`, `CustomEvent`, and `MouseEvent` seem to use both styles in this file, which makes me think this isn't related to Closure minification - maybe it has some effect in TS that I'm not recognizing?

packages/webcomponentsjs/ts_src/platform/custom-event.ts
```diff
@@ -0,0 +1,101 @@
...
+  }
+}
+
+// CustomEvent constructor shim
+if (!window.CustomEvent || isIE && (typeof window.CustomEvent !== 'function')) {
+  window['CustomEvent'] =
+      (<T extends unknown>(inType: string, params?: CustomEventInit<T>) => {
+        params = params || {};
+        const e =
+            /** @type {!CustomEvent} */ (document.createEvent('CustomEvent'));
```
bicknellr, line 70: Do we still need this Closure type?

packages/webcomponentsjs/ts_src/platform/promise.ts
```diff
@@ -0,0 +1,63 @@
...
+http://polymer.github.io/AUTHORS.txt The complete set of contributors may be
+found at http://polymer.github.io/CONTRIBUTORS.txt Code distributed by Google as
+part of the polymer project is also subject to an additional IP rights grant
+found at http://polymer.github.io/PATENTS.txt
+*/
+
+// No easy way to provide typings for a file imported this way from
+// node modules. So we'll just ignore the import and cast locally.
+// @ts-ignore
+import PromisePolyfillRaw from '../../node_modules/promise-polyfill/src/index.js';
```
bicknellr, line 15: The other file removed the `../../node_modules/` part, would that work here too? Also, does TS write out modified import specifiers itself or leave them in a Node-like format?

packages/webcomponentsjs/ts_src/platform/flag-parser.ts
```diff
@@ -0,0 +1,82 @@
...
+
+export {};
+
+interface ExtendedWindow extends Window {
+  WebComponents?: {flags?: Flags;};
+  ShadyDOM?: {force?: boolean|string; noPatch?: boolean | string;};
+  customElements: CustomElementRegistry&{
+    forcePolyfill?: string|boolean;
+  };
+}
```
bicknellr, line 20: Why this vs `declare namespace window { ... }`?

packages/webcomponentsjs/ts_src/platform/custom-event.ts
```diff
@@ -0,0 +1,101 @@
...
+ * Copyright (c) 2016 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt The complete set of authors may be found
+ * at http://polymer.github.io/AUTHORS.txt The complete set of contributors may
+ * be found at http://polymer.github.io/CONTRIBUTORS.txt Code distributed by
+ * Google as part of the polymer project is also subject to an additional IP
+ * rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+
+export {};
```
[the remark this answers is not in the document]
rictic: Yeah, it still has the script/module distinction

packages/webcomponentsjs/ts_src/platform/custom-event.ts
rictic: It's actually a closure type checking issue. Without the quotes, closure was noticing that we were assigning to these symbols and was broadening their types to such a degree that other unrelated code was failing to compile. Using quotes is making these symbols opaque to closure compiler

packages/webcomponentsjs/ts_src/platform/flag-parser.ts
rictic: `declare namespace window { ... }` makes a global change for all code in the entire compilation unit (which could be a lot of code in a big app that imports this).

This style keeps the changes isolated.

packages/webcomponentsjs/ts_src/platform/custom-event.ts
rictic: Nope, fixing

packages/webcomponentsjs/ts_src/platform/promise.ts
rictic: Looks like we can remove that part, yeah. I didn't mean to commit the version with it removed from the other file, but it looks like closure compiler can handle this style.

> Also, does TS write out modified import specifiers itself or leave them in a Node-like format?

TS will never modify import specifiers, it always passes them through untouched.

bicknellr: > TS will never modify import specifiers, it always passes them through untouched.

I was going to say "in that case, we probably shouldn't since this will change these packages from being web-compatible to requiring a build step" but it looks like the file in promise-polyfill that we link to wasn't web-compatible in the first place because the specifiers it uses internally assume node-style lookups.

rictic: Yeah.

There's also a perspective where all of this code is specifically for the oldest browsers, which don't have native modules anyways.

packages/webcomponentsjs/ts_src/platform/custom-event.ts
bicknellr: Huh, that sounds like a Closure bug; wouldn't the right way to handle `window.x = y;` be to fail if `y` doesn't match the expected type of `window.x`, rather than updating the type of `window.x`?

PR #299, opened 2020-04-24
Freshen all package-lock.json files

aomarks:
I just ran "npm run nuke".

This reduces the number of NPM security audit issues from 1991 to 21.

Most of the remaining ones seem to come from webcomponents-tester-legacy. We
should migrate to Karma anyway.


aomarks:
Yeah I think a bot that did this e.g. daily would be great. I think if it passes CI, we should just fully automate merging into master, otherwise it should create a PR for us to review/fix.

PR #300, opened 2020-04-24
Clean up pull request template and code owners

aomarks:


.github/CODEOWNERS
```diff
@@ -1,7 +1 @@
-packages/custom-elements/ @azakus @bicknellr
-packages/html-imports/ @azakus
-packages/shadycss/ @azakus
-packages/shadydom/ @azakus @sorvell
-packages/template/ @azakus
-packages/url/ @azakus
-packages/webcomponentsjs/ @azakus
```
bicknellr, line 7: I think @dfreedm and @sorvell should weigh in here.

aomarks: This change just means that we'll be the default automatic reviewers on new PRs, and I figured we'd add in other people to review as needed. Seems like it makes sense for us to do triage in this way. Anybody who wants to get notifications for PRs can just watch the repo. But I'm fine if @dfreedm @sorvell still also want to be assigned to every PR, their call.

dfreedm: Nope, I'm happy to let the people doing the work be the auto reviewers :)

sorvell: Same as dfreedm.

Issue #252, opened 2020-01-22
Polyfill CSS Shadow Parts (::part)


justinfagnani:
See https://developer.mozilla.org/en-US/docs/Web/CSS/::part

PR #306, opened 2020-04-24
Document ShadyCSS Shadow Parts design

aomarks:
This readme documents the proposal for Shadow Parts support in ShadyCSS, and what its limitations will be.

(Note this PR is merging into the `shady-parts` branch, which is where I will start sending PRs to build up Shadow Parts support for incremental code review. Once the feature is ready there, we'll have a final PR that merges `shady-parts` into `master`).

https://github.com/webcomponents/polyfills/issues/252

Fixes https://github.com/webcomponents/polyfills/issues/309

### DIFF
--- a/packages/shadycss/shadow-parts.md
+++ b/packages/shadycss/shadow-parts.md
@@ -99,10 +99,10 @@ represented as shown below, thus there is no "resetting" after the second rule
 has matched an ancestor:
 
 ```css
-x-a x-a.odd[shady-part~=x-a:x-a:num] {
+x-a x-a.odd[shady-part~="x-a:x-a:num"] {
   color: red;
 }
-x-a x-a.even[shady-part~=x-a:x-a:num] {
+x-a x-a.even[shady-part~="x-a:x-a:num"] {
   color: green;
 }
 ```
@@ -138,7 +138,7 @@ ShadyCSS:
 
 ```html
 <style>
-  x-a x-b [shady-part~=x-a:x-b:greeting] {
+  x-a x-b [shady-part~="x-a:x-b:greeting"] {
     color: red;
   }
 </style>
@@ -176,7 +176,7 @@ x-b::part(foo bar) {
 ShadyCSS:
 
 ```css
-x-a x-b [shady-part~=x-a:x-b:foo][shady-part~=x-a:x-b:bar] {
+x-a x-b [shady-part~="x-a:x-b:foo"][shady-part~="x-a:x-b:bar"] {
   color: red;
 }
 ```
@@ -222,10 +222,10 @@ ShadyCSS:
 
 ```html
 <style>
-  x-a x-b [shady-part~=x-a:x-b:salutation] {
+  x-a x-b [shady-part~="x-a:x-b:salutation"] {
     color: red;
   }
-  x-b x-c [shady-part~=x-b:x-c:greeting] {
+  x-b x-c [shady-part~="x-b:x-c:greeting"] {
     font-weight: bold;
   }
 </style>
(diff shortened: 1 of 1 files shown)


packages/shadycss/shadow-parts.md
```diff
@@ -0,0 +1,307 @@
...
+  <x-b>
+    <div shady-part="x-a:x-b:greeting">Hello</div>
+  </x-b>
+</x-a>
+```
+
+### Document styles
+
+As is standard in ShadyCSS, document-level styles must be opted-in with
+`ShadyCSS.CustomStyleInterface.addCustomStyle`. The first component of a shimmed
```
sorvell, line 156: When the polyfill bundle supports a version of `custom-style` replace `addCustomStyle` with that element.

packages/shadycss/shadow-parts.md
```diff
@@ -0,0 +1,307 @@
...
+```
+
+Workarounds will vary based on specific scenario. For the above example, a good
+workaround would be to use two unique part names: `numOdd`, `numEven`.
+
+## How it works
+
+The basic approach is to convert `::part` rules to attribute selectors that
+encode the _providing_ scope, the _receiving_ scope, and the part name. Each
+shadow host then discovers its parent scope, and applies the correct attribute
```
sorvell, line 117: Does this update when the element moves into a different shadowRoot? If not, suggest adding this as a limitation.

packages/shadycss/shadow-parts.md
```diff
@@ -0,0 +1,307 @@
...
+}
+x-a x-a.even[shady-part~=x-a:x-a:num] {
+  color: green;
+}
+```
+
+Workarounds will vary based on specific scenario. For the above example, a good
+workaround would be to use two unique part names: `numOdd`, `numEven`.
+
+## How it works
```
sorvell, line 113: I suggest adding a section on `Performance` or perhaps integrating that into this section. This would describe more specifically the type of work done, when it's done, and specifically focus on what work is done once v. per instance.

packages/shadycss/shadow-parts.md
```diff
@@ -0,0 +1,307 @@
...
+  <x-b>
+    <x-c>
+      <div shady-part="x-b:x-c:greeting
+                       x-a:x-b:salutation">Hello</div>
+    </x-c>
+  </x-b>
+</x-a>
+```
+
+### Custom properties
```
sorvell, line 243: Clarify that this is only needed on browsers that don't natively support css custom properties.

packages/shadycss/shadow-parts.md
aomarks: Done.

packages/shadycss/shadow-parts.md
```diff
@@ -0,0 +1,345 @@
...
+
+In addition, when at least one part node is found in a shadow root, the host is
+checked for an `exportparts` attribute. If found, then _its_ host is checked
+for an `exportparts` attribute, and so on, walking upwards through the
+hierarchy of shadow roots, to construct a part name forwarding map. This map is
+cached for each shadow root, so each shadow root is only walked once. This map
+is then used to add additional `shady-part` attributes to each applicable part
+node.
+
+If a `part` attribute is added or removed from an element, then the appropriate
```
sorvell, line 333: By what mechanism is this discovered?

aomarks: I added a special case to the shady DOM `Element.setAttribute` shim:

https://github.com/webcomponents/polyfills/blob/shady-parts-dev/packages/shadydom/src/patches/Element.js#L102

Issue #308, opened 2020-04-27
Polyfills interface library


aomarks:
Publish a new library `@webcomponents/interface` that provides an importable module that interacts with the global polyfills APIs (e.g. ShadyCSS's styleElement) if they are present, and no-ops otherwise. With TypeScript typings.

PR #313, opened 2020-04-29
Create new shadycss interface module

aomarks:
This new library provides an interface to the Web Components polyfills that is importable, type-safe, and easier to use than the `window.ShadyCSS` and similar polyfill globals. It's 371B Brotli compressed.

For example, to imperatively set a CSS custom property in a way that works both natively and with ShadyCSS:

**Before:**

```typescript
if ((window as any).ShadyCSS && !(window.any).ShadyCSS.nativeShadow) {
  window.ShadyCSS.styleSubtree(this, {'--my-color', 'red'});
} else {
  this.style.setProperty('--my-color', 'red');
}
```

**After:**

```typescript
import * as shadyCss from '@webcomponents/shadycss';

shadyCss.styleSubtree(this, {'--my-color', 'red'})
```

See the `README.md` for all the details.

https://github.com/webcomponents/polyfills/issues/308


packages/interface/src/interface.ts
```diff
@@ -0,0 +1,165 @@
...
+ * The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+
+interface PolyfillsWindow extends Window {
+  ShadyCSS?: ShadyCSS;
+}
+
+interface ShadyCSS {
```
rictic, line 15: would we be able to import these types if shadycss was in TypeScript?

justinfagnani: Only if this package had a dependency on the shady-css package

packages/interface/src/interface.ts
```diff
@@ -0,0 +1,165 @@
...
+  CustomStyleInterface?: CustomStyleInterface;
+}
+
+interface CustomStyleInterface {
+  addCustomStyle(style: HTMLStyleElement): void;
+}
+
+type CustomPropertyValues = { [property: string]: string };
+
+const polyfillsWindow = window as PolyfillsWindow;
```
justinfagnani, line 33: I think you can redeclare `window` rather than a new variable:

```ts
declare var window: PolyfillsWindow;
```

packages/interface/src/interface.ts
```diff
@@ -0,0 +1,165 @@
...
+/**
+ * Propagate custom properties from the given `document` level `<style>` tag.
+ * Changes occur asynchronously as a microtask.
+ *
+ * If ShadyCSS is not active, or if the `custom-style-interface` library has not
+ * been loaded, then this function does nothing.
+ *
+ * Note that the `custom-style-interface` library is not automatically loaded by
+ * the `loader` or `bundle` scripts. It must be loaded separately.
+ */
```
justinfagnani, line 155: Should we mention https://github.com/webcomponents/polyfills/tree/master/packages/shady-css-scoped-element?

packages/interface/tsconfig.json
```diff
@@ -0,0 +1,21 @@
+{
+  "compilerOptions": {
+    "target": "es2018",
```
justinfagnani, line 3: Set target to 2017 until we upgrade all of our targets.

packages/interface/tsconfig.json
```diff
@@ -0,0 +1,21 @@
...
+    "module": "es2015",
+    "moduleResolution": "node",
+    "strict": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "preserveConstEnums": true,
+    "forceConsistentCasingInFileNames": true,
+    "rootDir": "src/",
+    "outDir": "dist/",
```
justinfagnani, line 12: I'd prefer lib/ for consistency

packages/interface/tsconfig.json
```diff
@@ -0,0 +1,21 @@
...
+    "sourceMap": true,
+    "incremental": true,
+    "tsBuildInfoFile": "dist/.tsbuildinfo"
+  },
+  "lib": ["DOM"],
```
justinfagnani, line 17: this looks like it's in the wrong place. Should be in `compilerOptions`.

packages/interface/src/interface.ts
aomarks: > LGTM, but I'd like to talk just once more about what polyfills we might facade like this and whether this should be in shady-css... just to make sure we have it right before publishing

Yeah, I could be convinced either way.

Pros:

1. Maybe more logical organization. These are proxies for shadycss methods, so seems reasonable to expect that shadycss itself would provide it?

2. If we added more proxies for other WC polyfill methods here (e.g. shadydom), then some users will improt more code than they need. Many bundlers don't tree-shake within a module (though we're talking about tiny amounts of code either way). Also organizationally seems to make sense for each polyfill to provide its own functions (though the shadycss/shadydom distinction isn't all that meaningful to most users, I feel).

3. One less package for us to maintain and users to know about.

Cons:

1. NPM dependency bloat. It makes sense for libraries/components (e.g. lit-element, any re-usable component that e.g. sets a css property imperatively) to call these APIs by installing this dep, but they don't actually need the polyfill itself. No runtime cost of course, just install cost.

2. The optics of simplistic dependency analysis.

WDYT?

justinfagnani: I totally agree with your pros and cons. I'm just not 100% the conclusion I draw from them :)

packages/interface/src/interface.ts
aomarks: Yep, but that's not published yet (is on my near-term todo list).

On that topic, it seems like shady-css-scoped-element (and maybe this new interface module too?) should automatically work without the user having to load custom-style-interface themselves. custom-style-interface is pretty small, so I wonder if we should just bundle custom-style-interface with shadycss? Otherwise, it could be a dependency of shady-css-scoped-element, like it is in Polymer.

packages/interface/tsconfig.json
aomarks: Done.

packages/interface/tsconfig.json
aomarks: Done.

packages/interface/tsconfig.json
aomarks: Done.

packages/interface/src/interface.ts
aomarks: Done.

aomarks:
> LGTM, but I'd like to talk just once more about what polyfills we might facade like this and whether this should be in shady-css... just to make sure we have it right before publishing

(I don't know how my reply to this ended up on the *would we be able to import these types if shadycss was in TypeScript?* thread, but that's where I sent it by mistake).

packages/interface/src/interface.ts
aomarks: I'm leaning shadycss package, seems less weird.

rictic: IMO it's better for the types to be defined once, and for us to do `import type` here

aomarks: Sorry, I messed up somehow with this thread by replying in the wrong place. My first reply, and the next two comments, don't have anything to do with your original question and Justin's first reply. Interested in your opinion about where this package should live, not sure if your last comment is related to that.

aomarks:
PTAL @justinfagnani @rictic 

1. I moved the module directly into the shadycss package. You now import it like this:

    ```typescript
    import * as shadyCss from '@webcomponents/shadycss';
    ```

2. I removed some functions in the interest of making the API as simple as possible for now. We can easily add these back later, but it's harder to remove them.
    - `styleDocument`: It's trivial to write `styleElement(document.documentBody)` instead, which does the same thing.

    - `addCustomStyle`: I think a more ergonomic interface for most users will be to use `<shady-css-scoped>`. So rather than having two ways to do the same thing, I'm thinking we just support `<shady-css-scoped>` (it still needs to be published, see https://github.com/webcomponents/polyfills/issues/305).

    - `scopedStylesArePolyfilled` and `customPropertiesArePolyfilled`: Since the other APIs automatically check these, I'm not sure users will actually need to access these bits anymore.

3. Updated the existing shadycss documentation to use this new module.

4. Removed documentation about type extension elements, since they aren't suported natively, and I'm not sure we should be advertising the feature anymore.

aomarks:
I totally screwed up this PR by accidentally force pushing a shallow clone while I was playing around with something unrelated, such that GitHub thought this PR was closed, and now won't let me re-open it.

@justinfagnani @rictic please see https://github.com/webcomponents/polyfills/pull/323 for the next round of review.

Issue #54, opened 2018-10-07
Add CHANGELOGs to all packages


LarsDenBakker:
It would be useful to have a changelog for this library. It has bugs/integration issues quite often, which is somewhat understandable given the complexity of the polyfill, so I have to clamp the version. But when clamping the version, It's hard to tell what I'm losing out.

cekvenich:
I notice version 2.4 was release. But there is no change log. ?

PR #322, opened 2020-05-07
GitHub Action to require PRs to touch a CHANGELOG

aomarks:
Only runs on PRs targeting the master branch, and can be disabled by adding the "Skip Changelog" label.

Fixes https://github.com/webcomponents/polyfills/issues/319
Related to https://github.com/webcomponents/polyfills/issues/54

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -20,6 +20,10 @@ jobs:
 
     - name: Check for CHANGELOG changes
       run: |
+        # Only the latest commit of the feature branch is checkoued out
+        # automatically (see https://github.com/webcomponents/polyfills). To
+        # diff with the base branch, we need to fetch that too (and we only
+        # need its latest commit).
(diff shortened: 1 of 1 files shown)


.github/workflows/changelog.yml
```diff
@@ -0,0 +1,31 @@
...
+  changelog:
+    runs-on: ubuntu-latest
+    if: "!contains(github.event.pull_request.labels.*.name, 'Skip Changelog')"
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Check for CHANGELOG changes
+      run: |
+        git fetch origin ${{ github.base_ref }} --depth=1
```
bicknellr, line 23: Does this commit need to be fetched vs just checking it out?

aomarks: Yep, because the checkout action only gives us the latest commit on the feature branch by default. Added a comment.

PR #325, opened 2020-05-08
Add build:watch script to all packages

aomarks:
Watch and build all the packages with:

  `npm run build:watch`

Watch and build some specific packages with e.g.:

  `npm run build:watch -- --scope=@webcomponents/{shadycss,shadydom}`

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,6 @@
     "@gulp-sourcemaps/sources-content": "^1.0.0",
     "babel-core": "^6.26.3",
     "chokidar-cli": "^2.1.0",
-    "concurrently": "^5.2.0",
     "del": "^3.0.0",
     "eslint": "^5.9.0",
     "eslint-plugin-html": "^5.0.3",
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
   "scripts": {
     "bootstrap": "lerna bootstrap --ci --reject-cycles",
     "build": "lerna run build --stream",
-    "build:watch": "lerna run build:watch --stream",
+    "build:watch": "lerna run build:watch --parallel",
     "lint": "lerna run lint --stream",
     "test": "wct",
     "test-pack": "lerna exec \"npm pack\"",
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
   "scripts": {
     "bootstrap": "lerna bootstrap --ci --reject-cycles",
     "build": "lerna run build --stream",
+    "build:watch": "lerna run build:watch --stream",
     "lint": "lerna run lint --stream",
     "test": "wct",
     "test-pack": "lerna exec \"npm pack\"",
@@ -16,6 +17,8 @@
   "devDependencies": {
     "@gulp-sourcemaps/sources-content": "^1.0.0",
     "babel-core": "^6.26.3",
+    "chokidar-cli": "^2.1.0",
+    "concurrently": "^5.2.0",
     "del": "^3.0.0",
     "eslint": "^5.9.0",
     "eslint-plugin-html": "^5.0.3",
--- a/packages/html-imports/package.json
+++ b/packages/html-imports/package.json
@@ -22,7 +22,7 @@
   "license": "BSD-3-Clause",
   "scripts": {
     "build": "gulp",
-    "build:watch": "chokidar 'src/**/*.js' -c 'npm run build'",
+    "build:watch": "chokidar --initial 'src/**/*.js' -c 'npm run build'",
     "debug": "cp src/html-imports.js html-imports.min.js",
     "prepack": "npm run build"
   },
--- a/packages/html-imports/package.json
+++ b/packages/html-imports/package.json
@@ -22,6 +22,7 @@
   "license": "BSD-3-Clause",
   "scripts": {
     "build": "gulp",
+    "build:watch": "chokidar 'src/**/*.js' -c 'npm run build'",
     "debug": "cp src/html-imports.js html-imports.min.js",
     "prepack": "npm run build"
   },
--- a/packages/shady-css-scoped-element/package.json
+++ b/packages/shady-css-scoped-element/package.json
@@ -23,7 +23,7 @@
   "license": "BSD-3-Clause",
   "scripts": {
     "build": "gulp",
-    "build:watch": "chokidar 'src/**/*.js' -c 'npm run build'",
+    "build:watch": "chokidar --initial 'src/**/*.js' -c 'npm run build'",
     "debug": "cp src/shady-css-scoped-element.js shady-css-scoped-element.min.js",
     "prepack": "npm run build"
   },
--- a/packages/shady-css-scoped-element/package.json
+++ b/packages/shady-css-scoped-element/package.json
@@ -23,6 +23,7 @@
   "license": "BSD-3-Clause",
   "scripts": {
     "build": "gulp",
+    "build:watch": "chokidar 'src/**/*.js' -c 'npm run build'",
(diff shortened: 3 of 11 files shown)


package.json
```diff
@@ -16,6 +17,8 @@
   "devDependencies": {
     "@gulp-sourcemaps/sources-content": "^1.0.0",
     "babel-core": "^6.26.3",
+    "chokidar-cli": "^2.1.0",
+    "concurrently": "^5.2.0",
```
bicknellr, line 21: IMO we should put these in the individual polyfills too. Though, I think the repo is already in a bad state where some dev dependencies are explicitly listed in the individual polyfills' packages and some aren't.

package.json
```diff
@@ -7,6 +7,7 @@
   "scripts": {
     "bootstrap": "lerna bootstrap --ci --reject-cycles",
     "build": "lerna run build --stream",
+    "build:watch": "lerna run build:watch --stream",
```
bicknellr, line 10: I don't think this works for watching all of the repos since lerna is going to wait for a package's dependencies in the build graph to complete before running their dependents but none of the `build:watch` scripts ever actually completes on its own. For example,`webcomponentsjs` will never be rebuilt as a result of changes to one of the individual polyfills.

packages/html-imports/package.json
```diff
@@ -22,6 +22,7 @@
   "license": "BSD-3-Clause",
   "scripts": {
     "build": "gulp",
+    "build:watch": "chokidar 'src/**/*.js' -c 'npm run build'",
```
bicknellr, line 25: AFAICT, chokidar doesn't do an initial run like `tsc --watch`, so the ones that use chokidar to watch the actual source (vs JS output from `tsc` that are written during the build) probably need to be prefixed with `npm run build && `.

I found this [`ignoreInitial` option](https://github.com/paulmillr/chokidar#path-filtering) - it reads like it's talking about this situation and should do what we want by default, but I'm not seeing that happen in my repo. Maybe I'm missing something?

package.json
aomarks: Ah, good point, I hadn't tested webcomponentsjs (there aren't any other inter package dependencies I don't think, so it didn't really matter for those).

Switched from `--stream` to `--parallel`. The latter ignores dependencies (and they actually use a "watch" command as the example!): https://github.com/lerna/lerna/tree/master/commands/run#--parallel

Also extended the chokidar watched paths for webcomponentsjs to include the JS files of its dependencies.

packages/html-imports/package.json
aomarks: You want to look at https://github.com/kimmobrunfeldt/chokidar-cli for the flags, you might have been looking at the underlying library instead of the CLI. I added the `--initial` flag, which does this.

package.json
aomarks: Hmm, yeah I wasn't sure which pattern to follow. Here's a comment from one of the maintainers of Lerna suggesting putting shared `devDependencies` only at the top level: https://github.com/lerna/lerna/issues/258#issuecomment-229237660

I guess advantages of being only in the top-level are:
- Faster and smaller install.
- Keeps versions in sync.

Disadvantages:
- Can't run the command when you are cd'd into the specific package unless you manually set your path. (But maybe this reflecting the fact that you're supposed to run all your commands from the top-level using `--scope` when needed?).

Sound right? Other pros/cons? Would seem to favor putting more stuff in `devDependencies`?

There's also the `lerna link convert` command, which I hadn't looked at before: https://github.com/lerna/lerna#common-devdependencies. It makes everything a `devDependency` of the top-level, and adds `file:../` dependencies. I tried it out... but I'm not sure it's doing the right thing, it seems to do weird things, like removing all of the inter-package dependencies from `webcomponentsjs` (not even `file:` deps).

bicknellr: I could probably be swayed to convert this repo to have all dev dependencies at the top, so if you want to leave them here that works for me. Though, it kind of feels off because the dependencies aren't actually at their most specific point, but I guess it's not as if we're going to work on these packages in isolation anymore.

I didn't know about `lerna link convert` - after running it I can see what you're saying about it doing weird stuff. I think the removal of all the inter-package dev dependencies makes sense from the perspective of wanting to put all dev dependencies at the top but what confuses me is that it added all of the packages as regular dependencies of the top level package:

```
+  "dependencies": {
+    "@webcomponents/custom-elements": "file:packages/custom-elements",
+    "@webcomponents/html-imports": "file:packages/html-imports",
+    "@webcomponents/shady-css-scoped-element": "file:packages/shady-css-scoped-element",
+    "@webcomponents/shadycss": "file:packages/shadycss",
+    "@webcomponents/shadydom": "file:packages/shadydom",
+    "@webcomponents/template": "file:packages/template",
+    "@webcomponents/url": "file:packages/url",
+    "@webcomponents/webcomponentsjs": "file:packages/webcomponentsjs",
+    "tests": "file:packages/tests"
   }
```

Those were never dependencies of that package, so I'm not sure why it thinks they should be there? It also added them as dev dependencies (expected) but with `file:` refs that look broken as well as not attempting to link to the local webcomponentsjs package:

```
   "devDependencies": {
     "@gulp-sourcemaps/sources-content": "^1.0.0",
+    "@webcomponents/custom-elements": "file:../custom-elements",
+    "@webcomponents/html-imports": "file:../html-imports",
+    "@webcomponents/shadycss": "file:../shadycss",
+    "@webcomponents/shadydom": "file:../shadydom",
+    "@webcomponents/template": "file:../template",
+    "@webcomponents/url": "file:../url",
+    "@webcomponents/webcomponents-platform": "^1.0.0",
+    "@webcomponents/webcomponentsjs": "^2.4.2",
     "babel-core": "^6.26.3",
```

packages/shady-css-scoped-element/package.json
```diff
@@ -23,6 +23,7 @@
   "license": "BSD-3-Clause",
   "scripts": {
     "build": "gulp",
+    "build:watch": "chokidar 'src/**/*.js' -c 'npm run build'",
```
bicknellr, line 26:
```suggestion
    "build:watch": "chokidar --initial 'src/**/*.js' -c 'npm run build'",
```
(guessing based on the others)

packages/html-imports/package.json
bicknellr: SGTM

package.json
bicknellr: Ok, this seems to be building everything in the graph for me now. I think we should get the tests package included in there too though. Particularly, starting from a clean checkout, `npm i && npm run bootstrap && npm run build:watch` in one shell and `npm run test` in another should pass. I haven't investigated much of what to do to fix it but I tried running `npx lerna run build --scope=tests` right after all the other builds settled and tried again but got the same error.

package.json
aomarks: Ok, I'll leave it as-is, and we can look into making things more consistent and do a bit more research on the best approach later.

package.json
aomarks: Ah good catch. I didn't realize there was anything to build in the tests package. Fixed.

Tests are all passing for me locally.

bicknellr: I tried again and wasn't seeing them passing so I did a bit more digging and the custom elements polyfill didn't get built. I wasn't sure what was happening exactly so I removed concurrently and changed chokidar to watch the TS inputs and run `npm run build` instead and that seemed to fix it. I think there's some weird edge case where chokidar isn't recognizing that files were written by tsc.

aomarks: Interesting, maybe a linux vs mac thing? (I'm on linux). Removed concurrently for now.

Issue #311, opened 2020-04-28
Land initial implementation of shadow parts for shadycss in shady-parts branch


aomarks:

aomarks:
Closing in favor of smaller scoped issues.

PR #329, opened 2020-05-19
Primitive shadow parts support

aomarks:
Note: suggest reviewing the two commits separately, since the first one contains some reformatting.

This supports basic patterns of styling parts whenever a node is inserted into a shadow root. It also adds and tests a `disableShadowParts` ShadyCSS setting.

Support for more kinds of mutations, exportparts, CSS custom properties are coming in follow up PRs.

Part of https://github.com/webcomponents/polyfills/issues/311 and https://github.com/webcomponents/polyfills/issues/252. Fixes https://github.com/webcomponents/polyfills/issues/337.


packages/shadycss/src/style-transformer.js
```diff
@@ -315,6 +316,12 @@ class StyleTransformer {
           NTH, (m, type, inner) => `:${type}(${inner.replace(/\s/g, '')})`);
       selector = this._twiddleNthPlus(selector);
     }
+    if (!disableShadowParts && PART.test(selector)) {
+      // Hacky transform "::part(foo bar)" to "::part(foo,bar)" so that
+      // SIMPLE_SELECTOR_SEP isn't confused by the spaces.
+      // TODO(aomarks) Can we make SIMPLE_SELECTOR_SEP smarter instead?
```
sorvell, line 322: It would be nice to work this out, but acknowledge that it may be non-trivial.

packages/shadycss/src/style-transformer.js
```diff
@@ -396,6 +405,20 @@ class StyleTransformer {
     return output.join('');
   }
 
+  _transformPartSelector(selector, scope) {
```
sorvell, line 408: Doc with an example transformation.

packages/shadydom/src/patches/Node.js
```diff
@@ -375,6 +375,12 @@ export const NodePatches = utils.getOwnPropertyDescriptors({
     } else if (node.ownerDocument !== this.ownerDocument) {
       this.ownerDocument.adoptNode(node);
     }
+    if (!utils.disableShadowParts) {
```
sorvell, line 378: See previous comment about performance concerns and consider instead a tighter integration at https://github.com/webcomponents/polyfills/blob/master/packages/shadydom/src/patches/Node.js#L311-L333.

packages/shadycss/src/shadow-parts.js
```diff
@@ -165,3 +165,47 @@ export function formatShadyPartSelector(
...
+    // browsers that getRootNode is sometimes undefined?
+    return;
+  }
+  const root = parentNode.getRootNode();
+  if (root === document) {
+    // Parts in the document scope would never have any effect. Return early so
+    // we don't waste time querying it.
+    return;
+  }
+  const parts = parentNode.querySelectorAll('[part]');
```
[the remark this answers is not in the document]
sorvell: Also consider doing this in `prepareTemplateDom` so that only dynamic parts need to be changed here.

packages/shadydom/src/utils.js
```diff
@@ -12,6 +12,10 @@ import {shadyDataForNode} from './shady-data.js';
 /** @type {!Object} */
 export const settings = window['ShadyDOM'] || {};
 
+/** @type {boolean} */
+export const disableShadowParts =
+    Boolean(window['ShadyCSS'] && window['ShadyCSS']['disableShadowParts']);
```
sorvell, line 17: Shouldn't this be `true` if `ShadyCSS` does not exist?

packages/shadycss/src/shadow-parts.js
```diff
@@ -165,3 +165,47 @@ export function formatShadyPartSelector(
     })
     .join('');
 }
+
+/*  eslint-disable no-unused-vars */
+/**
+ * Perform any needed Shadow Parts updates after a ShadyDOM insertBefore call
```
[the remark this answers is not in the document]
sorvell: It would be much less stress if we didn't have to alter the part attribute. Can you describe how this would limit the fidelity of the shim?

packages/shadycss/src/shadow-parts.js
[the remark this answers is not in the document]
sorvell: This should be searching `newNode` and `newNode.querySelectorAll`, not `parentNode.querySelectorAll`.

packages/shadycss/src/shadow-parts.js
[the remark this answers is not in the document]
aomarks: Yeah, interesting. I thought of an example that seemed sort of realistic where your suggestion would have caused an error, but then I thought of a modification that resolves it:

(1)
```html
<style>
  /* error, the <div> grand-child of <x-other> shouldn't match */
  x-super .x-host[part~=foo] { color: red; }
  /* but is OK with this modification */
  x-super x-host.x-super .x-host[part~=foo] { color: red; }
</style>

<x-super>
  <x-host class="x-super">
    <div class="x-host" part="foo">
  <x-other class="x-super">
    <x-host class="x-other">
      <div class="x-host" part="foo">
```

However, you can still construct some patterns that fail:

(2)
```html
<style>
  x-super x-host.x-super .x-host[part~=foo] { color: red; }
</style>

<x-super>
  <x-host class="x-super">
    <div class="x-host" part="foo">
    <x-other class="x-super">
      <x-host class="x-other">
        <!-- error: should not be red -->
        <div class="x-host" part="foo">
```

Whereas with the `shady-parts` attribute, the selectors are much tighter:

(3)
```html
<style>
  x-super x-host [shady-part~="x-super:x-host:foo"] { color: red; }
</style>

<x-super>
  <x-host>
    <div class="x-host" part="foo" shady-part="x-super:x-host:foo">
    <x-other>
      <x-host>
        <!-- OK -->
        <div class="x-host" part="foo" shady-part="x-other:x-host:foo">
```

This is not to say that the `shady-parts` approach is perfect -- once you start repeating the same elements in a nesting hierarchy then you get similar problems (as described at https://github.com/webcomponents/polyfills/blob/shady-parts-basic/packages/shadycss/shadow-parts.md#3-recursion-is-crude). But it is at least a more limited set of patterns.

I guess the next question is, are examples like (2) above acceptable gaps in fidelity? It seems a little tricky to workaround the (2) example unless you can control the part names. WDYT? Can you think of any other selector tricks that would work even better?

--------------

The other consideration here is exported parts. In the implementation in https://github.com/webcomponents/polyfills/pull/331, the way we support export parts is by adding additional `shady-parts` attributes to each part node based on the `exportparts` attributes found when we walk up the tree of shadow hosts. This allows one transformed style rule to match parts in all possible export paths.

I guess the way to do this if we used this other approach would be: each time we discover that an exported part needs to receive some style, we clone that style rule (or update its selector) with the new scope. We'd need to do per-instance styling here, though, in cases where two instances of the same element have different exportparts attributes. I already interact with per-instance styling in https://github.com/webcomponents/polyfills/pull/333 to support custom properties within part rules, so it wouldn't be that wild. The difference here, though, would be that we'd be doing per-instance styling for more cases, even in Edge when native custom property support is available.

WDYT, am I thinking about this right?

packages/shadycss/src/scoping-shim.js
```diff
@@ -23,6 +23,8 @@ import templateMap from './template-map.js';
 import * as ApplyShimUtils from './apply-shim-utils.js';
 import {updateNativeProperties, detectMixin} from './common-utils.js';
 import {CustomStyleInterfaceInterface, CustomStyleProvider} from './custom-style-interface.js'; // eslint-disable-line no-unused-vars
+import * as shadowParts from './shadow-parts.js';
```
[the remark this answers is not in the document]
sorvell: I think you've evaluated it correctly.

packages/shadycss/src/shadow-parts.js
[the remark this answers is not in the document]
sorvell: You'd need to add support for knowing when to not support the querySelector, but there's precedent for this: https://github.com/webcomponents/polyfills/blob/master/packages/shadydom/src/patches/Node.js#L309.

packages/shadycss/src/shadow-parts.js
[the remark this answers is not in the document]
sorvell: Thanks for considering this so carefully. To summarize, nesting is a problem with any of our approaches, but with the current approach nesting of (A->B) inside an (A->B) is problematic whereas with the proposed approach nesting an A inside an A is problematic.

I was considering this without exportparts. I think considering that argues for how you've done it.

packages/shadycss/src/shadow-parts.js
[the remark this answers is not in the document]
aomarks: > Does this do a `querySelectorAll` on every call to `insertBefore`? If so, have you measured the performance impact of this? I'd expect it to be noticeable. ShadyDOM already conditionally does a walk during insertBefore (for example, to find slots); maybe consider a tighter integration to be able to participate in that walk.

No benchmarks yet. I will add them in an upcoming task: https://github.com/webcomponents/polyfills/issues/344

Will look into replacing the `querySelectorAll` with an integration into the ShadyDOM walk as an upcoming task: https://github.com/webcomponents/polyfills/issues/345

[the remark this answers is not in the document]
aomarks: > Also, it seems like this could cut down on the cost of the `querySelectorAll` since I guess I could store some kind of index of the part locations (or at the very least, a bit to tell me if it is even worth querying for parts), but I couldn't actually fully pre-process the template to add the `shady-part` attribute, since the value of that attribute depends on knowing the super-scope, which we don't know until we're connected. Does that sound right?

@sorvell any thoughts on this? I've filed https://github.com/webcomponents/polyfills/issues/346 for future investigation.

packages/shadycss/src/style-transformer.js
aomarks: Acknowledged. Leaving for now with TODO.

packages/shadycss/src/style-transformer.js
aomarks: Done.

packages/shadydom/src/patches/Node.js
aomarks: Acknowledged. Added TODO and filed https://github.com/webcomponents/polyfills/issues/345 for upcoming task.

packages/tests/shadycss/shadow-parts/insertions.html
```diff
@@ -0,0 +1,146 @@
...
+      xa = document.createElement('x-a');
+      document.body.appendChild(xa);
+      xb = pierce(xa, 'x-b');
+    });
+
+    teardown(() => {
+      document.body.removeChild(xa);
+    });
+
+    test('appendChild: part', () => {
```
sorvell, line 57: Add test for appending/inserting a Text node?

packages/shadycss/src/shadow-parts.js
```diff
@@ -189,11 +191,34 @@ export function onInsertBefore(parentNode, newNode, referenceNode) {
...
+  // querySelectorAll calls here by integrating the part check into the walk
+  // that ShadyDOM already does to find slots.
+  // https://github.com/webcomponents/polyfills/issues/345
+  if (newNode instanceof DocumentFragment) {
+    // E.g. a template stamp. Note this call occurs before the native insert, so
+    // DocumentFragment children are still contained by newNode.
+    parts = newNode.querySelectorAll('[part]');
+  } else if (newNode instanceof Text) {
+    // E.g. an innerHTML assignment.
+    parts = parentNode.querySelectorAll('[part]');
```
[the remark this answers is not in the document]
aomarks: Good question, I got muddled in some debugging and was thinking the `innerHTML` call would show up as an insert with text, but obviously that's wrong. Removed.

packages/tests/shadycss/shadow-parts/insertions.html
aomarks: Good call, added test and an early return so that we don't crash in this case.

packages/tests/shadycss/shadow-parts/insertions.html
```diff
@@ -0,0 +1,163 @@
...
+    test('insertBefore: DocumentFragment with part', () => {
+      const fragment = new DocumentFragment();
+      const pa = document.createElement('div');
+      pa.setAttribute('part', 'pa');
+      fragment.appendChild(pa);
+      xb.shadowRoot.insertBefore(fragment, null);
+      assert.equal(color(pa), red);
+    });
+
+    test('appendChild: text node', () => {
```
sorvell, line 112:
```suggestion
    test('insertBefore: text node', () => {
```

aomarks: Done.

PR #332, opened 2020-05-21
Manually dispatch events for connected, unassigned targets.

### LINKED ISSUE #95: [ShadyDOM] Cannot listen for events from unslotted Light DOM children
### Description

Events dispatched from Light DOM children cannot be handled unless they are distributed in a `slot`.

#### Live Demo

https://stackblitz.com/edit/shady-dom-events-no-slot

#### Steps to Reproduce

1. Create two elements and nest one in the other.
   * Do not add a `<slot>`
1. Have the child dispatch an event
1. Have the parent listen for that event

#### Expected Results
The event should be handled by the parent.

#### Actual Results
In polyfilled browsers the parent element never receives the event.

### Browsers Affected

- [ ] Chrome
- [ ] Firefox
- [x] Edge
- [ ] Safari 9
- [ ] Safari 8
- [x] IE 11

### Versions
- webcomponents: v2.2.7

bicknellr:
This PR causes Shady DOM to manually dispatch all events with targets that are disconnected in the real tree. This causes events to propagate correctly if their path is split between connected and disconnected trees (in the real tree), which can happen if the target (or one of its ancestors) is connected but not assigned to a slot (in the user-facing tree).

~~- [ ] Merge Polymer/polymer#5663~~ (Assuming this passes internal tests, we're considering this low-risk enough to not require that Polymer's `...-changed` events go through this path.)

Fixes #95.

### DIFF
--- a/packages/shadydom/src/patch-events.js
+++ b/packages/shadydom/src/patch-events.js
@@ -306,50 +306,50 @@ function fireHandlers(event, node, phase) {
 }
 
 function shadyDispatchEvent(e) {
-  let path = e.composedPath();
-  let node;
+  const path = e.composedPath();
+  const retargetedPath = path.map(node => retarget(node, path));
+  const bubbles = e.bubbles;
+
+  let currentTarget;
   // override `currentTarget` to let patched `target` calculate correctly
   Object.defineProperty(e, 'currentTarget', {
     get: function() {
-      return node;
+      return currentTarget;
     },
     configurable: true
   });
+
+  let eventPhase = Event.CAPTURING_PHASE;
+  Object.defineProperty(e, 'eventPhase', {
+    get: function() {
+      return eventPhase;
+    },
+  });
+
   for (let i = path.length - 1; i >= 0; i--) {
-    node = path[i];
+    currentTarget = path[i];
+    eventPhase = currentTarget === retargetedPath[i] ? Event.AT_TARGET : Event.CAPTURING_PHASE;
     // capture phase fires all capture handlers
-    fireHandlers(e, node, 'capture');
+    fireHandlers(e, currentTarget, 'capture');
     if (e.__propagationStopped) {
       return;
     }
   }
 
-  let eventPhase = Event.AT_TARGET;
-  Object.defineProperty(e, 'eventPhase', {get() { return eventPhase; }});
-
-  const bubbles = e.bubbles;
-  // the event only needs to be fired when owner roots change when iterating the event path
-  // keep track of the last seen owner root
-  let lastFiredRoot;
   for (let i = 0; i < path.length; i++) {
-    node = path[i];
-    const nodeData = shadyDataForNode(node);
-    const root = nodeData && nodeData.root;
-    // Listeners on the deepest node in the path and the deepest node in each
-    // root of all nodes in the path should be called in the AT_TARGET phase.
-    const atDeepestNodeInRoot = i === 0 || (root && root === lastFiredRoot);
-    if (bubbles || atDeepestNodeInRoot) {
-      eventPhase = atDeepestNodeInRoot ? Event.AT_TARGET : Event.BUBBLING_PHASE;
-      fireHandlers(e, node, 'bubble');
-      // don't bother with window, it doesn't have `getRootNode` and will be last in the path anyway
-      if (node !== window) {
-        lastFiredRoot = node[utils.SHADY_PREFIX + 'getRootNode']();
-      }
+    currentTarget = path[i];
+    const atTarget = currentTarget === retargetedPath[i];
+    if (atTarget || bubbles) {
+      eventPhase = atTarget ? Event.AT_TARGET : Event.BUBBLING_PHASE;
+      fireHandlers(e, currentTarget, 'bubble');
       if (e.__propagationStopped) {
         return;
       }
     }
   }
+
+  eventPhase = Event.NONE;
+  currentTarget = null;
 }
 
 function listenerSettingsEqual(savedListener, node, type, capture, once, passive) {
--- a/packages/shadydom/src/patch-events.js
+++ b/packages/shadydom/src/patch-events.js
@@ -396,6 +396,9 @@ function targetNeedsPathCheck(node) {
 /** @this {Node} */
 export function dispatchEvent(event) {
   flush();
+  // If the target is disconnected from the real document, it might still be
+  // connected in the user-facing tree. To allow its path to potentially
+  // include both connected and disconnected parts, dispatch it manually.
   if (this instanceof Node && !utils.documentContains(document, this)) {
     if (!event['__target']) {
       patchEvent(event, this);
--- a/packages/shadydom/src/patch-events.js
+++ b/packages/shadydom/src/patch-events.js
@@ -335,17 +335,18 @@ function shadyDispatchEvent(e) {
     node = path[i];
     const nodeData = shadyDataForNode(node);
     const root = nodeData && nodeData.root;
-    if (!(i === 0 || (root && root === lastFiredRoot)) && !bubbles) {
-      continue;
-    }
-
-    fireHandlers(e, node, 'bubble');
-    // don't bother with window, it doesn't have `getRootNode` and will be last in the path anyway
-    if (node !== window) {
-      lastFiredRoot = node[utils.SHADY_PREFIX + 'getRootNode']();
-    }
-    if (e.__propagationStopped) {
-      return;
+    // Listeners on the deepest node in the path and the deepest node in each
+    // root of all nodes in the path should be called in the AT_TARGET phase.
+    const isRetargetedTarget = i === 0 || (root && root === lastFiredRoot);
+    if (bubbles || isRetargetedTarget) {
+      fireHandlers(e, node, 'bubble');
+      // don't bother with window, it doesn't have `getRootNode` and will be last in the path anyway
+      if (node !== window) {
+        lastFiredRoot = node[utils.SHADY_PREFIX + 'getRootNode']();
+      }
+      if (e.__propagationStopped) {
+        return;
+      }
     }
   }
 }
--- a/packages/shadydom/src/patch-events.js
+++ b/packages/shadydom/src/patch-events.js
@@ -396,7 +396,7 @@ function targetNeedsPathCheck(node) {
 /** @this {Node} */
 export function dispatchEvent(event) {
   flush();
-  if (this instanceof Node && !(this === document || document.documentElement[utils.NATIVE_PREFIX + 'contains'](this))) {
+  if (this instanceof Node && !utils.documentContains(document, this)) {
     if (!event['__target']) {
       patchEvent(event, this);
     }
--- a/packages/shadydom/src/patch-events.js
+++ b/packages/shadydom/src/patch-events.js
@@ -305,7 +305,7 @@ function fireHandlers(event, node, phase) {
   }
 }
 
-function retargetNonBubblingEvent(e) {
+function shadyDispatchEvent(e) {
   let path = e.composedPath();
   let node;
   // override `currentTarget` to let patched `target` calculate correctly
@@ -327,6 +327,7 @@ function retargetNonBubblingEvent(e) {
   // set the event phase to `AT_TARGET` as in spec
   Object.defineProperty(e, 'eventPhase', {get() { return Event.AT_TARGET }});
 
+  const bubbles = e.bubbles;
   // the event only needs to be fired when owner roots change when iterating the event path
   // keep track of the last seen owner root
   let lastFiredRoot;
@@ -334,15 +335,17 @@ function retargetNonBubblingEvent(e) {
     node = path[i];
     const nodeData = shadyDataForNode(node);
     const root = nodeData && nodeData.root;
-    if (i === 0 || (root && root === lastFiredRoot)) {
-      fireHandlers(e, node, 'bubble');
-      // don't bother with window, it doesn't have `getRootNode` and will be last in the path anyway
-      if (node !== window) {
-        lastFiredRoot = node[utils.SHADY_PREFIX + 'getRootNode']();
-      }
-      if (e.__propagationStopped) {
-        return;
-      }
+    if (!(i === 0 || (root && root === lastFiredRoot)) && !bubbles) {
+      continue;
+    }
+
+    fireHandlers(e, node, 'bubble');
+    // don't bother with window, it doesn't have `getRootNode` and will be last in the path anyway
+    if (node !== window) {
+      lastFiredRoot = node[utils.SHADY_PREFIX + 'getRootNode']();
+    }
+    if (e.__propagationStopped) {
+      return;
     }
   }
 }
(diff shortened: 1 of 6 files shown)


packages/shadydom/src/patch-events.js
```diff
@@ -391,7 +394,14 @@ function targetNeedsPathCheck(node) {
 /** @this {Node} */
 export function dispatchEvent(event) {
   flush();
-  return this[utils.NATIVE_PREFIX + 'dispatchEvent'](event);
+  if (this instanceof Node && !document.documentElement[utils.NATIVE_PREFIX + 'contains'](this)) {
```
bicknellr, line 397: The `this instanceof Node` check here is not something I tried in the perf tests: I hadn't considered non-Node event targets and Node's `contains` throws if you pass it a non-Node.

packages/shadydom/src/patch-events.js
```diff
@@ -305,7 +305,7 @@ function fireHandlers(event, node, phase) {
   }
 }
 
-function retargetNonBubblingEvent(e) {
+function shadyDispatchEvent(e) {
```
bicknellr, line 308: Maybe needs bikeshedding, the non-bubbling part of `retargetNonBubblingEvent` isn't correct anymore though.

packages/shadydom/src/patch-events.js
```diff
@@ -508,12 +518,12 @@ export function addEventListener(type, fnOrObj, optionsOrCapture) {
...
-    this.__handlers[type] = this.__handlers[type] ||
-      {'capture': [], 'bubble': []};
-    this.__handlers[type][capture ? 'capture' : 'bubble'].push(wrapperFn);
-  } else {
+  this.__handlers = this.__handlers || {};
+  this.__handlers[type] = this.__handlers[type] ||
+    {'capture': [], 'bubble': []};
+  this.__handlers[type][capture ? 'capture' : 'bubble'].push(wrapperFn);
+
+  if (!nonBubblingEventsToRetarget[type]) {
```
bicknellr, line 526: These changes cause _all_ event listeners to be manually added to `__handlers` so that `shadyDispatchEvent` can run them itself. Before, it was just things in `nonBubblingEventsToRetarget` (`focus`, `blur`).

sorvell: This was unexpected for me since I know we have to keep track of the handlers for `removeEventListener`. However, there we keep a map of wrapper function to node/options and here we need the reverse map from node to wrapper function.

packages/tests/shadydom/event-path.html
```diff
@@ -930,6 +937,22 @@
       w.dispatchEvent(new Event('foo'));
       assert.isTrue(heardEvent);
     });
+
+    test('events dispatched to unassigned slottables bubble through the tree above them', () => {
```
sorvell, line 941: Let's include a test like the example in this issue: https://github.com/Polymer/lit-element/issues/658.

packages/shadydom/src/patch-events.js
```diff
@@ -327,22 +327,25 @@ function retargetNonBubblingEvent(e) {
...
-    if (i === 0 || (root && root === lastFiredRoot)) {
-      fireHandlers(e, node, 'bubble');
-      // don't bother with window, it doesn't have `getRootNode` and will be last in the path anyway
-      if (node !== window) {
-        lastFiredRoot = node[utils.SHADY_PREFIX + 'getRootNode']();
-      }
-      if (e.__propagationStopped) {
-        return;
-      }
+    if (!(i === 0 || (root && root === lastFiredRoot)) && !bubbles) {
```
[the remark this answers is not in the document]
bicknellr: This is meant to handle composed, non-bubbling events: it's deciding if it should call listeners on `node` using `root && root === lastFiredRoot` to figure out if `node` hosts the last shadow root in which a listener was called, which ends up calling listeners on only the deepest nodes in each shadow root in the path (and `i === 0` handles when `node` is the real target).

This section is confusing and took me a while to realize what was happening when I initially came across it. (Especially this condition, which I haven't helped.) I'll try to rearrange it to make it clearer.

packages/shadydom/src/patch-events.js
```diff
@@ -391,7 +394,14 @@ function targetNeedsPathCheck(node) {
 /** @this {Node} */
 export function dispatchEvent(event) {
   flush();
-  return this[utils.NATIVE_PREFIX + 'dispatchEvent'](event);
+  if (this instanceof Node && !(this === document || document.documentElement[utils.NATIVE_PREFIX + 'contains'](this))) {
```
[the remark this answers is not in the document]
dfreedm: `document.contains` is an O(1) check at least in Chrome, vs a O(height of tree) for other nodes

packages/shadydom/src/patch-events.js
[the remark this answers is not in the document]
bicknellr: This has a bug: it always should fire if `i === 0`.

packages/shadydom/src/patch-events.js
```diff
@@ -572,8 +582,8 @@ const EventPatchesDescriptors = utils.getOwnPropertyDescriptors(EventPatches);
 const SHADY_PROTO = '__shady_patchedProto';
 const SHADY_SOURCE_PROTO = '__shady_sourceProto';
 
-function patchEvent(event) {
-  event['__target'] = event.target;
+function patchEvent(event, forcedTarget = undefined) {
+  event['__target'] = forcedTarget !== undefined ? forcedTarget : event.target;
```
bicknellr, line 586:
```suggestion
function patchEvent(event, target = event.target) {
  event['__target'] = target;
```
I need to double check that parameter defaults can refer to earlier parameters but this would be clearer if so.

bicknellr: Added.

packages/shadydom/src/patch-events.js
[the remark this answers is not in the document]
bicknellr: Ok, this check now runs through a function that tries `document.contains` before `document.documentElement.contains`. Should I still benchmark this?

packages/shadydom/src/patch-events.js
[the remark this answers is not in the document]
bicknellr: Updated, PTAL

packages/tests/shadydom/event-path.html
bicknellr: I still need to do this.

packages/shadydom/src/patch-events.js
```diff
@@ -324,17 +324,22 @@ function retargetNonBubblingEvent(e) {
...
   for (let i = 0; i < path.length; i++) {
     node = path[i];
     const nodeData = shadyDataForNode(node);
     const root = nodeData && nodeData.root;
-    if (i === 0 || (root && root === lastFiredRoot)) {
+    // Listeners on the deepest node in the path and the deepest node in each
+    // root of all nodes in the path should be called in the AT_TARGET phase.
+    const atDeepestNodeInRoot = i === 0 || (root && root === lastFiredRoot);
+    if (bubbles || atDeepestNodeInRoot) {
+      eventPhase = atDeepestNodeInRoot ? Event.AT_TARGET : Event.BUBBLING_PHASE;
```
bicknellr, line 342: This needs to happen during capturing also...

packages/tests/shadydom/event-path.html
bicknellr: I added another test that I think is closer to what's happening in that bug. I think the reason I was struggling to come up with a scenario that worked for this without including LitElement itself in the tests was that LitElement actually schedules its first render asynchronously after the first `connectedCallback` but I was under the impression that it was sync during `connectedCallback`. The test I added includes a parent and child, where the parent creates an empty shadow root for itself in its constructor and the child dispatches the event in its constructor (tighter timing than the example in the bug), but I simplified by making the parent never actually fill the shadow root with any content. In LitElement, this happens late enough that I think it doesn't strike me as useful to include in the test: the `connectedCallback` dispatching the event happens sync inside an `innerHTML` set in the test (which is all we care about observing) so there's no need to actually make the child appear in the real tree, AFAICT.

packages/shadydom/src/patch-events.js
bicknellr: This is fixed.

packages/tests/shadydom/event-path.html
```diff
@@ -930,6 +937,123 @@
...
+            ShadyDOM.wrap(this).addEventListener('some-event', () => {
+              this.listenerCallCount++;
+            });
+          }
+        });
+
+        customElements.define('child-dispatcher-element', class extends HTMLElement {
+          constructor() {
+            super();
+            ShadyDOM.wrap(this).dispatchEvent(new Event('some-event', {bubbles: true}));
```
bicknellr, line 1038: Using `wrap` instead of `wrapIfNeeded` here was necessary to make this test pass. Should I add `dispatchEvent` to the notes for `wrapIfNeeded`?

https://github.com/webcomponents/polyfills/blob/de5c30c3a51e0890bbecf6dc44b5e8364201730a/packages/shadydom/src/shadydom.js#L83-L93

packages/shadydom/src/patch-events.js
bicknellr: We talked about this in a meeting and decided `this instanceof Node` shouldn't be a perf problem and there's another thread about `document.contains` vs `document.documentElement.contains` below so I'm going to close this one.

packages/shadydom/src/patch-events.js
```diff
@@ -306,48 +306,50 @@ function fireHandlers(event, node, phase) {
...
-    fireHandlers(e, node, 'bubble');
-    // don't bother with window, it doesn't have `getRootNode` and will be last in the path anyway
-    if (node !== window) {
-      lastFiredRoot = node[utils.SHADY_PREFIX + 'getRootNode']();
-    }
-    if (e.__propagationStopped) {
-      return;
+    currentTarget = path[i];
+    const atTarget = currentTarget === retargetedPath[i];
+    if (atTarget || bubbles) {
```
sorvell, line 342: Can this loop be broken out of when `atTarget` and `bubbles` are false?

packages/shadydom/src/utils.js
```diff
@@ -92,7 +92,14 @@ export const microtask = (callback) => {
   twiddle.textContent = content++;
 }
 
-export const hasDocumentContains = Boolean(document.contains);
+const hasDocumentContains = Boolean(document.contains);
+
+/** @type {function(!Document, !Node): boolean} */
+export const documentContains = (doc, node) => (
+  doc === node ||
+  (hasDocumentContains && doc[NATIVE_PREFIX + 'contains'](node)) ||
```
sorvell, line 100: Suggest doing this check once and writing different `documentContains` implementations based on the result.

packages/tests/shadydom/event-path.html
```diff
@@ -938,20 +938,121 @@
...
+        ShadyDOM.wrap(innerNoSlot).appendChild(target);
+
+        assert.strictEqual(ShadyDOM.wrap(target).assignedSlot, null);
+
+        const eventLog = [];
+        const pushToEventLog = (e) => {
+          eventLog.push({currentTarget: e.currentTarget, eventPhase: e.eventPhase});
+        };
+
+        // `ShadyDOM.wrapIfNeeded(outerNoSlot).shadowRoot` will also see this event in its path,
```
sorvell, line 988: Do we have enough information to correct this?

packages/tests/shadydom/event-path.html
sorvell: Yes, that sounds good.

packages/shadydom/src/patch-events.js
sorvell: Got it.

packages/tests/shadydom/event-path.html
```diff
@@ -930,6 +937,123 @@
       w.dispatchEvent(new Event('foo'));
       assert.isTrue(heardEvent);
     });
+
+    suite('unassigned slottable targets', () => {
```
sorvell, line 941: Suggest adding tests that stress this code path (if we don't have them) for:
1. events that bubble v. not
2. events that are composed v. not
3. events are not sent to removed event listeners

packages/shadydom/src/utils.js
bicknellr: Yeah, that's good idea. Also, I just realized that I was trying to be too clever here with the short-circuiting: if `document.contains` and `document.documentElement` both exist but `doc` doesn't contain `node`, then it will end up running both `contains` calls.

packages/shadydom/src/patch-events.js
bicknellr: Not always: if the event is non-bubbling, composed, and dispatched to an element in a shadow root, then both the real target and the shadow root host will see `atTarget` as true even though there are other items in `path` between them.

packages/tests/shadydom/event-path.html
bicknellr: Added. Also, good call, the listeners weren't being removed properly.

packages/shadydom/src/utils.js
bicknellr: Added.

packages/tests/shadydom/event-path.html
bicknellr: I updated it so that it swaps descriptors for `eventPhase` (similar to existing code for `currentTarget`) when a listener on a shadow root is called. AFAICT, Shady DOM currently throws when dispatching to a shadow root itself, so I didn't handle that case. Also, the order is still wrong / depends on the order the listeners were added because listeners on shadow roots are actually added to their hosts, so they're all part of the same list. Fixing that seems like it's more effort than it's worth here / out of scope for this PR given most events go through native `dispatchEvent`.

packages/tests/shadydom/event-path.html
```diff
@@ -999,17 +1125,25 @@
...
+          {currentTarget: ShadyDOM.wrap(fixtureRoot).shadowRoot, eventPhase: Event.CAPTURING_PHASE},
           {currentTarget: outerDiv, eventPhase: Event.CAPTURING_PHASE},
           {currentTarget: innerNoSlot, eventPhase: Event.CAPTURING_PHASE},
           {currentTarget: target, eventPhase: Event.AT_TARGET},
           {currentTarget: target, eventPhase: Event.AT_TARGET},
           {currentTarget: innerNoSlot, eventPhase: Event.BUBBLING_PHASE},
           {currentTarget: outerDiv, eventPhase: Event.BUBBLING_PHASE},
-          {currentTarget: outerNoSlot, eventPhase: Event.AT_TARGET},
+          {currentTarget: fixtureRoot, eventPhase: Event.AT_TARGET},
+          // The bubbling event listener on this shadow root is incorrectly run
```
sorvell, line 1137: It's unclear why the order of running the event listeners should be wrong. Can you explain that?

bicknellr: When you add an event listener to a shadow root, the wrapper actually adds a (wrapped) listener to the shadow root's host:

https://github.com/webcomponents/polyfills/blob/9bd1fc0e13ba3c68d691ff7115397b6e735a8edc/packages/shadydom/src/patches/ShadowRoot.js#L21-L30

This way, even though Shady DOM's shadow roots aren't nodes in the real tree, their listeners will be still called by natively dispatched events. However, the order in which they're added ends up affecting when they're called because all the listeners added by the user to the host or its shadow root are all added to that host's single native list of listeners. For example, if bubbling listener A is first added to some host and then bubbling listener B is added to its shadow root, a bubbling, composed event dispatched to a node inside the shadow root will call A and then B because A was added first, even though B would normally be called first in a browser with native Shadow DOM.

sorvell: Ah, got it, that is not ideal but makes sense.

PR #333, opened 2020-05-21
Support shadow parts consuming custom properties

aomarks:
When a `::part` rule consumes a custom property, the value of that property is determined at each scope where that rule is _received_, not where the rule was defined.

This means that, unless we have native custom properties (which already work fine here with no changes), any time we find a part node that is receiving a style that consumes a custom property, we need to switch that part's host to per-instance styling, just as though it was consuming a custom property in its own styles.

There is still a known issue where property values set in the exact same rule that they are consumed are not supported. I included a skipped test for this and will address in a follow up, since I need to do some more digging to find the fix.

Part of https://github.com/webcomponents/polyfills/issues/252 and https://github.com/webcomponents/polyfills/issues/311. Fixes https://github.com/webcomponents/polyfills/issues/338.

### DIFF
--- a/packages/shadycss/src/shadow-parts.js
+++ b/packages/shadycss/src/shadow-parts.js
@@ -471,7 +471,7 @@ function rescopeForCustomProperties(host, newStyleNodes, newProperties) {
  * A map from "providerScope:receiverScope" compound string key to an array of
  * ::part rules that were found during template preparation.
  *
- * We only populate this map if a ::part rule consumes a custom property and we
+ * We only populate this map if a ::part rule consumes a custom property and 
  * native custom properties are not available.
  *
  * This map is used to quickly lookup whether a given ::part rule being applied
--- a/packages/shadycss/src/shadow-parts.js
+++ b/packages/shadycss/src/shadow-parts.js
@@ -456,13 +456,13 @@ function rescopeForCustomProperties(host, newStyleNodes, newProperties) {
     // Inject this ::part rule into the array of style rules managed by this
     // instance. This way, the standard ShadyCSS mechanism for evaluating
     // per-instance styles will take effect.
-    if (!styleInfo.styleRules.rules.includes(styleNode)) {
+    if (styleInfo.styleRules.rules.indexOf(styleNode) === -1) {
       styleInfo.styleRules.rules.push(styleNode);
     }
 
     // We need to explicitly enumerate the active part rules on the styleInfo,
     // because that is used as part of the per-instance style cache lookup.
-    if (!styleInfo.customPropertyPartRules.includes(styleNode)) {
+    if (styleInfo.customPropertyPartRules.indexOf(styleNode) === -1) {
       styleInfo.customPropertyPartRules.push(styleNode);
     }
   }
@@ -470,7 +470,7 @@ function rescopeForCustomProperties(host, newStyleNodes, newProperties) {
   // Augment the array of CSS custom properties consumed by this instance,
   // because that is also used in the per-instance style cache lookup.
   for (const property of newProperties) {
-    if (!styleInfo.ownStylePropertyNames.includes(property)) {
+    if (styleInfo.ownStylePropertyNames.indexOf(property) === -1) {
       styleInfo.ownStylePropertyNames.push(property);
     }
   }
@@ -547,8 +547,8 @@ export function analyzeTemplatePartRules(providerScope, styleAst) {
   }
   for (const styleNode of styleAst.rules) {
     const selector = styleNode['selector'];
-    if (!selector || !selector.includes('::part')) {
-      // TODO(aomarks) We do the `includes` check here to make the case where no
+    if (!selector || selector.indexOf('::part') === -1) {
+      // TODO(aomarks) We do the `indexOf` check here to make the case where no
       // `::part` rules are used is as fast as possible, but we could get away
       // with just the `parsePartSelector` call we're doing next if the
       // difference is negligible. Needs benchmarking.
@@ -608,7 +608,7 @@ function lookupMatchingPartRules(scopeKey, partNames) {
   for (const entry of entries) {
     let matches = true;
     for (const partName of entry.partNames) {
-      if (!partNames.includes(partName)) {
+      if (partNames.indexOf(partName) === -1) {
         matches = false;
         break;
       }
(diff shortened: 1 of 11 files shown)


packages/shadycss/src/shadow-parts.js
```diff
@@ -339,9 +363,212 @@ export function onInsertBefore(parentNode, newNode, referenceNode) {
...
+    if (!selector || selector.indexOf('::part') === -1) {
+      continue;
+    }
+    const consumedProperties = findConsumedCustomProperties(
+      styleNode['cssText']
+    );
+    if (consumedProperties.length === 0) {
+      continue;
+    }
+    const parsed = parsePartSelector(selector);
```
sorvell, line 516: I think it's likely that parsing the selector is cheaper than finding consumed properties, so do that first?

packages/shadycss/src/style-properties.js
```diff
@@ -495,9 +495,20 @@ class StyleProperties {
     let scope = '.' + scopeId;
     let parts = StyleUtil.splitSelectorList(selector);
     for (let i=0, l=parts.length, p; (i<l) && (p=parts[i]); i++) {
-      parts[i] = p.match(hostRx) ?
-        p.replace(hostSelector, scope) :
-        scope + ' ' + p;
+      if (p.match(hostRx)) {
+        parts[i] = p.replace(hostSelector, scope);
```
sorvell, line 499: Suggest changing the name `parts` to something else since it's too easily confused with shadow part.

packages/shadycss/src/shadow-parts.js
```diff
@@ -325,11 +331,29 @@ export function onInsertBefore(parentNode, newNode, referenceNode) {
     superRoot === document ? 'document' : superRoot.host.localName;
   const exportMap = findExportedPartMappings(host);
 
+  // Things to keep track of that determine if and how we need to update this
```
sorvell, line 334: It's unclear why we need to do runtime work here to support custom properties. Need some explanation of the design. In particular, there's already a runtime system for shimming custom properties, why isn't that sufficient?

aomarks: Hopefully this explains how I'm thinking about it. Starting from what ShadyCSS already does:

1) Template is prepared. Custom properties consumed in this element's `<style>` are discovered.

2) An instance is connected and `styleElement` is called.

3) If the element uses custom properties, their values are computed for this instance. The style cache is checked given those property values, and a new `<style>` is created if there isn't already a suitable one.

Note that at this point in time, we don't know anything about part rules for this instance. We've only dealt with the custom properties that were directly declared in our template styles. It's not immediately obvious how to teach the existing code paths under `styleElement` to handle part styles, because at the point at which it runs, we don't even know what parts the instance has.

4) The instance DOM is first stamped.

5) We check the DOM for parts. We figure out where our host is in the DOM.

6) We check to see if this instance is receiving any part rules that consume a custom property. If so, we have a problem. The values of those properties could be different for each instance, because their values correspond to where the part style is _consumed_, not where it is defined. So our simple-case transformed part rules would have the wrong values.

7) We "adopt" these part rules that have custom properties by adding them to our instance's style info, and we re-run the custom property shim logic that does per-instance styling (the same function as step 3).

Does that help?

One thing that stands out to me from this description is that there is repeated work happening in (3) and (7). There's probably some optimization that could be done where we skip (3) if there's good reason to think that (7) could happen (e.g. because in our ancestry there is at least one part rule that consumes a custom property, even if we don't exactly know where it's going to get applied yet).

WDYT?

sorvell: Yes, that helps. I suggest making sure that (6) is clearly called out in comments in the code.

packages/shadycss/src/shadow-parts.js
```diff
@@ -339,9 +363,212 @@ export function onInsertBefore(parentNode, newNode, referenceNode) {
...
+ *     template this is.
+ * @param {!StyleNode} styleAst Parsed CSS for this template.
+ */
+export function analyzeTemplatePartRules(providerScope, styleAst) {
+  if (nativeCssVariables || !styleAst.rules) {
+    return;
+  }
+  for (const styleNode of styleAst.rules) {
+    const selector = styleNode['selector'];
+    if (!selector || selector.indexOf('::part') === -1) {
```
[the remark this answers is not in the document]
aomarks: I was thinking that because this is going to run for every template, I want the cost for users who using part styles at all to be as small as possible. I assumed the regexp would be slower than a regular substring search, although maybe the difference is so negligible that it's better to just re-use the regex, since the tradeoff is that if there *is* a part style, the regex would have sufficed?

Probably needs a benchmark huh? Added a TODO for a benchmark.

(Also I updated `indexOf` to `includes` since it's more readable).

packages/shadycss/src/shadow-parts.js
aomarks: Done.

packages/shadycss/src/style-properties.js
aomarks: Done (went with "segments").

packages/shadycss/src/shadow-parts.js
aomarks: Fleshed out the documentation of `rescopeForCustomProperties` with a better explanation. Also added a TODO and filed https://github.com/webcomponents/polyfills/issues/348 to think about the fact that we are calling `_applyStyleProperties` twice, in effect.

packages/shadycss/src/shadow-parts.js
```diff
@@ -359,9 +383,239 @@ export function onInsertBefore(parentNode, newNode, referenceNode) {
...
+function lookupMatchingPartRules(scopeKey, partNames) {
+  const entries = partRulesMap.get(scopeKey);
+  if (entries === undefined) {
+    return [];
+  }
+  const matching = [];
+  for (const entry of entries) {
+    let matches = true;
+    for (const partName of entry.partNames) {
+      if (partNames.indexOf(partName) === -1) {
```
sorvell, line 611: Not following this. Why is it ok to bail on a match if one of entry partNames is not in the partNames argument?

aomarks: Sorry, wasn't too clearly written. I renamed the function, and added some documentation.

What's happening here is that, we have some part node with some list of part names from the `part` attribute, and we're trying to figure out whether there are any `::part` rules that match it, so that we can do the special handling for custom properties. By the spec, the specifier `::part(foo bar)` only matches part names with both `foo` AND `bar`. So the bail here is because we need to check if the given part node has ALL of the parts specified in the `::part` specifier, and we bail as soon as one is missing.

PR #350, opened 2020-06-08
Compare node with ref_node before executing `replaceChild`

### LINKED ISSUE #349: [ShadyDOM] Node.prototype.replaceChild removes node when same ref_node is used
### Description

When using `Node.prototype.replaceChild` with the same node as ref_node, the node gets cleaned. (eg. removed from its parent) 

### Example

https://jsbin.com/qajapuxupa/1/edit?html,console,output

### Steps to reproduce

1. Create a `div` and a `span` within
2. Get a reference to both elements.
3. Execute `replaceChild` on `div`, but use `span` for both the node as the ref_node
4. The `span` element is not connected anymore.

#### Expected behavior

The `span` element is untouched. (check other browsers)

#### Actual behavior

The `span` element has been cleaned

### Version

@webcomponents/webcomponentsjs@2.4.3

### Browsers affected

- [ ] Chrome
- [ ] Firefox
- [x] Edge (non Chromium versions)
- [ ] Safari
- [x] IE 11

tmsns:
Fixes #349

### DIFF
--- a/packages/shadydom/CHANGELOG.md
+++ b/packages/shadydom/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Compare node with ref_node before executing `replaceChild`.
+  ([#350](https://github.com/webcomponents/polyfills/pull/350))
 - Remove outdated references to `customElements.nativeHTMLElement`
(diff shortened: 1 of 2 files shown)


tmsns:
@googlebot I signed it!

tmsns:
Hey @aomarks, I see you assigned yourself to this PR. Can I help with anything? 😊

Issue #351, opened 2020-06-08
Improve Shady Parts specificity by making part rules important


aomarks:
First pass: make all part rules `!important`

PR #358, opened 2020-06-11
Transform ::part rule declarations to !important

aomarks:
As described in the [design doc](https://github.com/webcomponents/polyfills/blob/shady-parts/packages/shadycss/shadow-parts.md#2-specificity-is-crude), transform all `::part` rule declarations to `!important`.

We do this to all `::part` rules as a crude way to improve the specificity behavior of shadow parts for the (hopefully) common simple use-case of a part node that has some local default styles that should be overridable by some `::part` rule. For example:

```html
<x-parent>
   #shadow-root
     <!-- This color should have precendence. Unless we transform it to
          !important, though it would not beat out the #foo selector
          from the <x-child> default part styles. -->
     <style>
       x-child::part(foo) { color: red; }
     </style>

     <x-child>
       #shadow-root
         <style>
           #foo { color: green; }
         </style>
         <div id="foo" part="foo">Hello</div>
```
 
Since this is all we do to help `::part` rule specificity, there are many patterns that still won't behave correctly. For example, adding a grand-parent scope to the above case:

```html
 <x-grandparent>
   #shadow-root
     <!-- This color should have precedence over both the <x-parent>
          and <x-child> ones, but we don't guarantee this. -->
     <style>
       x-parent::part(foo) { color: blue; }
     </style>
     <x-parent>...</x-parent>
```

This is because the 3 transformed selectors will look as below, so precedence is determined by template processing order:

```css
x-grandparent x-parent [shady-part~="x-grandparent:x-parent:foo"] {}
x-parent      x-child  [shady-part~="x-parent:x-child:foo"] {}
.x-child#foo {}
```
To workaround specificity issues like this, the user will need to manually increase the weight of rules that they want to take precedence (e.g. adding an `#id` selector to the `<x-grandparent>` rule).

Fixes https://github.com/webcomponents/polyfills/issues/351
Part of https://github.com/webcomponents/polyfills/issues/252

### DIFF
--- a/packages/shadycss/src/shadow-parts.js
+++ b/packages/shadycss/src/shadow-parts.js
@@ -824,3 +824,65 @@ function findPartRulesThatMatchPartNode(scopeKey, partNames) {
   }
   return matching;
 }
+
+/**
+ * Transform all declarations in the given ::part rule to be !important.
+ *
+ * We do this to all ::part rules as a crude way to improve the specificity
+ * behavior of shadow parts for the (hopefully) common simple use-case of a
+ * part node that has some local default styles that should be overridable by
+ * some ::part rule. For example:
+ *
+ * <x-parent>
+ *   #shadow-root
+ *     <!-- This color should have precendence. Unless we transform it to
+ *          !important, though it would not beat out the #foo selector
+ *          from the <x-child> default part styles. -->
+ *     <style>
+ *       x-child::part(foo) { color: red; }
+ *     </style>
+ *
+ *     <x-child>
+ *       #shadow-root
+ *         <style>
+ *           #foo { color: green; }
+ *         </style>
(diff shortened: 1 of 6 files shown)


aomarks:
> The code here seems fine, but my overall concern is that I'm not sure this is worth doing. It generally seems much easier to restrict the type of styling that is allowed when you're defining parts, and less easy to explain the type of restrictions needed when styling parts. Unfortunately, this PR helps with the former and not the later.
> 
> I think it may make more sense to just document that you should only style parts via `[part=...]`. That seems like it will be generally compatible with the approach taken in the polyfill.
> 
> Then I think we need to focus more on the 'styling parts' issues here and make sure we have good tests and styling recommendations.

That's reasonable, I'll drop this change.

PR #363, opened 2020-07-10
formdata event: Add support for legacy browsers.

bicknellr:


packages/formdata-event/ts_src/wrappers/wrap_constructor.ts
```diff
@@ -0,0 +1,21 @@
+export const wrapConstructor = (Wrapper: any, Constructor: any, prototype: any) => {
```
aomarks, line 1: Missing license

aomarks, line 1: can you use `Object` or `unknown` or something else a little more specific instead of `any` here?

bicknellr: Added.

bicknellr: Ok, I expanded these; is there a built-in type for constructors in TypeScript?

aomarks: Nope, the typical one I see defined all over the place is: `type Constructor<T> = new (...args: any[]) => T;`

bicknellr: Huh, I tried that one originally but tsc kept giving me this error:
```
error TS2536: Type '"prototype"' cannot be used to index type 'C'.
```
But I just tried it again - inlining `Constructor<T>` for `C` this time - and it seems to work. Maybe the `extends` part in `C extends Constructor<T>` is related to that error? Anyway, I changed it since that definition seems more standard.

PR #364, opened 2020-07-13
Add element.dataset to list of unpatched custom-elements APIs

aomarks:


packages/custom-elements/README.md
```diff
@@ -66,6 +66,7 @@ and [HTML](https://html.spec.whatwg.org/) specifications are marked with the
     - `DOMTokenList` (`element.classList`)
     - `NamedNodeMap` (`element.attributes`)
     - `Attr` (`element.attributes.getNamedItem('attr-name')`)
+    - `element.dataset`
```
bicknellr, line 69: This list was just meant for excluded things in the DOM spec but `dataset` is part of HTML; what if it was mentioned as an example at the end of the bullet above that talks about HTML?

PR #369, opened 2020-07-28
Move part node discovery into Shady DOM

aomarks:
Before this change, any mutation to a shimmed shadow root triggered a `querySelectorAll` for `part` nodes that might be contained by it.

Now, we instead piggy-back on the existing walk that Shady DOM already does on mutations (e.g. to look for slots), and look for `part` attributes there.

This gives us a 10-13% performance improvement in a benchmark (also in this PR) where we insert 1000 elements, where each element provides some part rule to a child.

Screen Shot 2020-07-28 at 1 43 17 PM

Fixes https://github.com/webcomponents/polyfills/issues/345

### DIFF
--- a/packages/shadydom/src/patches/Node.js
+++ b/packages/shadydom/src/patches/Node.js
@@ -309,21 +309,21 @@ export const NodePatches = utils.getOwnPropertyDescriptors({
     const needsSlotFinding = ownerRoot && !node['__noInsertionPoint'] &&
         (!preferPerformance || node.nodeType === Node.DOCUMENT_FRAGMENT_NODE);
 
-    const partNodes = [];
-    const needsPartFinding =
+    let partNodes, exportpartsNodes;
+    if (
       // Only look for parts if there are known ::part rules active on the page,
       // and if shadow parts support hasn't been forced off.
       shadowPartsActive() &&
       // ownerRoot is undefined if we're inserting into the main document, and
       // host is undefined if we're not connected. Part rules can't apply in
       // either case.
-      ownerRoot && ownerRoot.host &&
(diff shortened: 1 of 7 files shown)


packages/shadydom/src/patches/Node.js
```diff
@@ -308,7 +308,33 @@ export const NodePatches = utils.getOwnPropertyDescriptors({
         !currentScopeIsCorrect(node, newScopeName);
     const needsSlotFinding = ownerRoot && !node['__noInsertionPoint'] &&
         (!preferPerformance || node.nodeType === Node.DOCUMENT_FRAGMENT_NODE);
-    if (needsSlotFinding || needsScoping) {
+
```
sorvell, line 311: This function is getting unwieldy. Think it'd improve readability to factor all the `needs-*` flags into separate functions and move the comments there.

aomarks (commented):
> It would be great to add a benchmark that stresses adding and removing nodes in general and to/from shadowRoots to help show that this code has minimal impact. It'd be good to test both when `shadowPartsActive()` is true and false.

I added a benchmark along those lines in the previous PR (https://github.com/webcomponents/polyfills/pull/368), which adds/removes a node (not a part) from a shadow root 1000 times. Here are the results for this PR (latest NPM release vs this PR vs this PR with shadow parts force disabled).

image

So, there is a chance it's as much as 2% slower than NPM on this benchmark. Maybe this could be further optimized by getting rid of some of the hopping through the shim to access the "have any part nodes been seen" variable.

sorvell (approved):
LGTM. Think the perf and other responses are reasonable.

PR #370, opened 2020-07-31
formdata: Capture browser APIs used in `dispatchFormdataForSubmission`.

bicknellr:
In #361 I wrote `dispatchFormdataForSubmission` using the DOM/HTML API directly from each object. This PR captures any new API used in that function in `environment/` and `environment_api/` and uses those instead. This helps prevent the polyfill from triggering behavior of other polyfills that are added after it.

### DIFF
--- a/packages/formdata-event/ts_src/environment_api/document.ts
+++ b/packages/formdata-event/ts_src/environment_api/document.ts
@@ -11,10 +11,10 @@
 
 import {methods as DocumentMethods} from '../environment/document.js';
 
-export const createElement = (
+export const createElement = <K extends keyof HTMLElementTagNameMap>(
   doc: Document,
-  localName: string,
+  localName: K,
   options?: ElementCreationOptions,
-): Element => {
-  return DocumentMethods.createElement.call(doc, localName, options);
+): HTMLElementTagNameMap[K] => {
+  return DocumentMethods.createElement.call(doc, localName, options) as HTMLElementTagNameMap[K];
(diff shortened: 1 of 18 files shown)


packages/formdata-event/ts_src/environment_api/document.ts
```diff
@@ -0,0 +1,20 @@
...
+ * http://polymer.github.io/LICENSE.txt The complete set of authors may be
+ * found at http://polymer.github.io/AUTHORS.txt The complete set of
+ * contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt Code
+ * distributed by Google as part of the polymer project is also subject to an
+ * additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+
+import {methods as DocumentMethods} from '../environment/document.js';
+
+export const createElement = (
```
aomarks, line 14: You should be able to preserve the `localName` -> return type mapping with something like this:

```ts
export const createElement = <K extends keyof HTMLElementTagNameMap>(
  doc: Document,
  localName: K,
  options?: ElementCreationOptions,
): HTMLElementTagNameMap[K] => {
    return document.createElement.call(doc, localName, options) as HTMLElementTagNameMap[K];
};
```

And then you won't need the cast later on when you do `createElement('input')`.

bicknellr: Very neat - added!

PR #378, opened 2020-08-24
Use Object.defineProperty instead direct define when create polyfilled HTMLCollection

louielin:
Direct define would cause enumerable issue. This behavior may break some js lib.
```js
Object.keys(document.querySelectorAll('body'));
// 0, item, namedItem

var buttons = document.querySelectorAll('button');
Object.keys(buttons).forEach(function(i) {
    var button = buttons[i];
    button.addEventListener(/*...*/); // raise error when "i" is "item" or "namedItem"
);
```
Using [Object.defineProperty](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty) and ensure `enumerable` equal to `false` will let "item" and "namedItem" methods disappear  in `Object.keys` and for loop


louielin:
> @googlebot I signed it!

@googlebot I signed it!

PR #379, opened 2020-08-27
Handle `stopPropagation` and `stopImmediatePropagation` calls on the 'submit' event.

bicknellr:
In the polyfill's normal flow, the 'formdata' event is dispatched when a 'submit' event dispatched to a form propagates up to a bubbling listener on the root containing the form. However, if a user calls `stopPropagation` or `stopImmediatePropagation` on the 'submit' event, then that event will never reach the listener at the root. To work around this problem, this PR wraps all 'submit' event listeners and checks after each listener if the 'formdata' event should be dispatched based on whether or not these methods have been called and if the wrapped listener would be the last to be called for that event.

### DIFF
--- a/packages/formdata-event/ts_src/event_listener_array.ts
+++ b/packages/formdata-event/ts_src/event_listener_array.ts
@@ -34,6 +34,30 @@ export class EventListenerArray {
 
   get bubblingCount() { return this._listeners.filter(record => !record.capture).length; }
 
+  get lastCapturingCallback() {
+    const listeners = this._listeners;
+    for (let i = listeners.length - 1; i >= 0; i--) {
+      const item = listeners[i];
+      if (item.capture) {
+        return item.callback;
+      }
+    }
+
+    return undefined;
+  }
+
+  get lastBubblingCallback() {
+    const listeners = this._listeners;
+    for (let i = listeners.length - 1; i >= 0; i--) {
+      const item = listeners[i];
+      if (!item.capture) {
+        return item.callback;
+      }
+    }
+
+    return undefined;
+  }
+
   push(record: EventListenerRecord) {
     const {callback, capture} = record;
 
--- a/packages/formdata-event/ts_src/event_listener_array.ts
+++ b/packages/formdata-event/ts_src/event_listener_array.ts
@@ -17,7 +17,7 @@ interface EventListenerRecord {
 /**
  * EventListenerArray keeps track of an array of event listeners, including
  * enough information to determine if they would be deduplicated by the browser:
- * type (always 'formdata' here), the callback itself, and capture flag. See
+ * type, the callback itself, and capture flag. See
  * https://dom.spec.whatwg.org/#add-an-event-listener for a full description of
  * the deduplicating behavior.
  */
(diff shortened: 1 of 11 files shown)


packages/formdata-event/ts_src/event_listener_array.ts
```diff
@@ -0,0 +1,87 @@
...
+
+interface EventListenerRecord {
+  readonly callback: EventListenerOrEventListenerObject;
+  readonly capture: boolean;
+}
+
+/**
+ * EventListenerArray keeps track of an array of event listeners, including
+ * enough information to determine if they would be deduplicated by the browser:
+ * type (always 'formdata' here), the callback itself, and capture flag. See
```
bicknellr, line 20: `(always 'formdata' here)` needs to be removed.

packages/formdata-event/ts_src/event_listener_array.ts
```diff
@@ -0,0 +1,87 @@
...
+ * https://dom.spec.whatwg.org/#add-an-event-listener for a full description of
+ * the deduplicating behavior.
+ */
+export class EventListenerArray {
```
bicknellr, line 24: The rest of this class could use some documentation.

packages/formdata-event/ts_src/formdata_listener_added.ts
```diff
@@ -132,21 +130,78 @@ export const submitCallback = (capturingEvent: Event) => {
 
   const shallowRoot = getRootNode(target);
 
+  // The wrapper generated here will handle dispatching the event if it bubbles
+  // all the way to `shallowRoot`.
```
bicknellr, line 134:
```suggestion
  // The wrapper generated here will handle dispatching a 'formdata' event if the
  // 'submit' event bubbles all the way to `shallowRoot`.
```

packages/formdata-event/ts_src/formdata_listener_added.ts
```diff
@@ -132,21 +130,78 @@ export const submitCallback = (capturingEvent: Event) => {
...
   // node of the tree containing the target form.
-  addEventListener.call(shallowRoot, 'submit', function bubblingCallback(bubblingEvent: Event) {
-    // Ignore any other 'submit' events that might bubble to this root.
-    if (bubblingEvent !== capturingEvent) {
-      return;
+  addEventListener.call(shallowRoot, 'submit', bubblingCallback);
+  submitListenerAdded(shallowRoot, bubblingCallback);
+};
+
+export const wrapSubmitListener = (listener: EventListenerOrEventListenerObject): EventListener => {
```
bicknellr, line 147: This needs documentation.

packages/formdata-event/ts_src/formdata_listener_added.ts
```diff
@@ -132,21 +130,78 @@ export const submitCallback = (capturingEvent: Event) => {
...
-    removeEventListener.call(shallowRoot, 'submit', bubblingCallback);
+    return result;
+  };
+};
 
-    // Ignore any cancelled events.
-    if (getDefaultPrevented(bubblingEvent)) {
-      return;
-    }
+const maybeDispatchFormdataForEvent = (e: Event) => {
```
bicknellr, line 192: This needs documentation.

aomarks (approved):
Gnarly problem!

PR #385, opened 2020-10-03
Add new entrypoints to webcomponentsjs for the 'platform' polyfills.

### LINKED ISSUE #386: Add entrypoints for 'platform' polyfills.
Some of the tests still reference the @webcomponents/webcomponents-platform package, but we merged that package into @webcomponents/webcomponentsjs and no longer publish it as a separate package. This prevents us from testing new 'platform' polyfills and makes it difficult to otherwise use them individually.

bicknellr:
Fixes #386.

I think the names for these bundles need some bikeshedding. Having these file names in the package is probably going to be very confusing for consumers:
```
bundles/webcomponents-js.js
bundles/webcomponents-dom.js
src/entrypoints/webcomponents-js-index.js
src/entrypoints/webcomponents-dom-index.js
```
`pf` is used in the other bundle names for 'platform' but now that they're split into two, we need some new separate name for each. `js` and `dom` were just the most obvious, but clearly aren't helpful if you're just looking around at the files in this package.


aomarks (approved):
Looks fine, I don't have an opinion about the names.

Can you remind me why we need to make these entrypoints for lit next (or update PR description)? I don't remember the discussion.

bicknellr:
Shady DOM and the custom elements polyfill both use the @webcomponents/webcomponents-platform package in [some of their tests](https://github.com/webcomponents/polyfills/blob/8b40e817e96d60ea9f1a13cf7a766e51f0e96644/packages/tests/shadydom/shady.html#L25) but we no longer have @webcomponents/webcomponents-platform as a distinct package in our source after merging all of the polyfill repos, so we can't update them (without recreating the package). These entrypoints will give us a place to put the newer polyfills that we're adding for Lit (`append`, `remove`, etc.) so that we can use them without importing the entire webcomponentsjs bundle in the tests.

kevinpschaaf:
`bundles/webcomponents-es6.js` (to disambiguate from webcomponents.js) and
`bundles/webcomponents-dom.js`

?

kevinpschaaf:
Or
`bundles/webcomponents-es6-shims.js`
`bundles/webcomponents-dom-shims.js`

bicknellr:
I think I might just go with the `pf_` prefixing thing mentioned in chat since IDK if these are going to be strictly ES6.

bicknellr:
Ok, updated, anyone opposed?

PR #388, opened 2020-10-06
Update the html-imports branch.

bicknellr:
This merges master into html-imports and updates the entrypoints that only exist in the html-imports branch to use the new platform bundles. https://github.com/webcomponents/polyfills/pull/388/commits/3dbf84797803c91faf68e8dc1eea7ba1f20badbf is the only new commit.


packages/formdata-event/CHANGELOG.md
```diff
@@ -0,0 +1,21 @@
+# Changelog
```
aomarks, line 1: These formdata-event files shouldn't be here, right?

aomarks: Nevermind, just realized this is a huge merge too!

aomarks (approved):

packages/formdata-event/CHANGELOG.md
bicknellr: Yeah, they're already in master but they show up in this diff since we haven't merged master into html-imports since that work was started.

PR #389, opened 2020-10-07
Add polyfills for select ParentNode APIs.

bicknellr:
Adds polyfills for `ParentNode`'s `append`, `prepend`, and `removeChildren` functions and wrappers to support these in Shady DOM.


packages/shadydom/src/patch-native.js
```diff
@@ -94,8 +94,10 @@ const ParentNodeAccessors = [
 
 const ParentNodeMethods = [
   'querySelector',
-  'querySelectorAll'
-  // 'append', 'prepend'
+  'querySelectorAll',
+  'append',
+  'prepend',
+  'replaceChildren',
```
bicknellr, line 100: Was this the right thing to do here? I'm not totally sure what this list is about. There seems to be some stuff later in this file that implies that they're on the wrong prototype in older Safari and that this is just about putting them in the correct spot - should I consider doing this for ChildNode in #390 too?

packages/shadydom/src/patches/ParentNode.js
```diff
@@ -93,7 +93,36 @@ export const ParentNodePatches = utils.getOwnPropertyDescriptors({
...
+  /** @this {Element} */
+  ['replaceChildren'](...args) {
+    let child;
+    while ((child = this[utils.SHADY_PREFIX + 'firstChild']) !== null) {
+      this[utils.SHADY_PREFIX + 'removeChild'](child);
+    }
+    for (const arg of args) {
+      const newChild = typeof arg === 'string' ? document.createTextNode(arg) : arg;
+      this[utils.SHADY_PREFIX + 'insertBefore'](newChild);
+    }
```
bicknellr, line 124: I should check if this is the right order (remove existing children, then insert new children), given that it probably feeds into MutationObserver.

bicknellr: This looks like the correct order: [`replaceChildren`](https://dom.spec.whatwg.org/#dom-parentnode-replacechildren) calls ['replace all'](https://dom.spec.whatwg.org/#concept-node-replace-all), which removes first and inserts second.

bicknellr:
Another issue regarding MutationObserver: these functions are meant to create one entry, but this implementation probably creates many. Is that something we care about?

bicknellr:
So following the spec more closely actually results in this being less work. The ['convert nodes into a node'](https://dom.spec.whatwg.org/#converting-nodes-into-a-node) algorithm basically just takes a list of nodes and puts them all in a fragment if there's not exactly one item, which can then be passed as one item to Shady DOM's `insertBefore`. This doesn't solve the multiple `MutationObserver` entries problem for `replaceChildren` though.

packages/tests/webcomponentsjs_/parent-node/replace-children.html
```diff
@@ -0,0 +1,92 @@
...
+<html>
+<head>
+  <title>ParentNode replaceChildren</title>
+  <script>
+    for (const item of [
+      Document.prototype,
+      DocumentFragment.prototype,
+      Element.prototype,
+    ]) {
+      delete item.replaceChildren;
```
sorvell, line 20: Why is this needed?

packages/tests/webcomponentsjs_/parent-node/replace-children.html
```diff
@@ -0,0 +1,92 @@
...
+    }
+  </script>
+  <script src="../../node_modules/@webcomponents/webcomponentsjs/bundles/webcomponents-pf_dom.js"></script>
+  <script src="../wct-config.js"></script>
+  <script src="../../node_modules/wct-browser-legacy/browser.js"></script>
+</head>
+<body>
+  <script>
+    const testAppend = (createContainer) => {
```
sorvell, line 29: testReplaceWith

packages/tests/webcomponentsjs_/parent-node/replace-children.html
```diff
@@ -0,0 +1,92 @@
...
+      test('Appends an element', () => {
```
sorvell, line 30: Titles of these tests should say "ReplaceWith"

packages/webcomponentsjs/ts_src/platform/parent-node/append.ts
```diff
@@ -0,0 +1,37 @@
+/**
```
sorvell, line 1: Can you explain why these APIs are being added here rather than just in shadydom?

sorvell (changes requested):

packages/shadydom/src/utils.js
```diff
@@ -230,3 +230,17 @@ export const assign = (target, source) => {
 export const arrayFrom = (object) => {
   return [].slice.call(/** @type {IArrayLike} */(object));
 };
+
+// Implements 'convert nodes into a node'.
+// https://dom.spec.whatwg.org/#converting-nodes-into-a-node
+export const convertNodesIntoANode = (...args) => {
+  if (args.length === 1) {
```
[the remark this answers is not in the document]
bicknellr: Good catch, should be fixed and tested now.

packages/tests/webcomponentsjs_/parent-node/append.html
```diff
@@ -0,0 +1,95 @@
...
+<html>
+<head>
+  <title>ParentNode append</title>
+  <script>
+    for (const item of [
+      Document.prototype,
+      DocumentFragment.prototype,
+      Element.prototype,
+    ]) {
+      delete item.append;
```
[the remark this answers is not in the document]
bicknellr: This and the others are here to remove the built-in implementations so that we can test it in browsers that actually do support it. I could also remove it if you think it's better to just run the tests against the built-in implementation, WDYT?

packages/tests/webcomponentsjs_/parent-node/append.html
```diff
@@ -0,0 +1,95 @@
...
+    }
+  </script>
+  <script src="../../node_modules/@webcomponents/webcomponentsjs/bundles/webcomponents-pf_dom.js"></script>
+  <script src="../wct-config.js"></script>
+  <script src="../../node_modules/wct-browser-legacy/browser.js"></script>
+</head>
+<body>
+  <script>
+    const testAppend = (createContainer) => {
```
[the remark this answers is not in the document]
bicknellr: Ok, I inlined these.

packages/tests/webcomponentsjs_/parent-node/prepend.html
```diff
@@ -0,0 +1,95 @@
...
+      delete item.prepend;
+    }
+  </script>
+  <script src="../../node_modules/@webcomponents/webcomponentsjs/bundles/webcomponents-pf_dom.js"></script>
+  <script src="../wct-config.js"></script>
+  <script src="../../node_modules/wct-browser-legacy/browser.js"></script>
+</head>
+<body>
+  <script>
+    const testAppend = (createContainer) => {
```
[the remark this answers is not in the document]
bicknellr: (Removed based on another comment in your review.)

packages/tests/webcomponentsjs_/parent-node/prepend.html
```diff
@@ -0,0 +1,95 @@
...
+      test('Appends an element', () => {
```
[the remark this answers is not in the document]
bicknellr: Fixed.

packages/tests/webcomponentsjs_/parent-node/replace-children.html
bicknellr: (replied in earlier comment)

packages/tests/webcomponentsjs_/parent-node/replace-children.html
bicknellr: (Removed based on another comment in your review.)

packages/tests/webcomponentsjs_/parent-node/replace-children.html
bicknellr: Fixed.

packages/webcomponentsjs/ts_src/platform/parent-node/append.ts
bicknellr: This is to separate the polyfill for the feature itself from the part that adds support for the feature to Shady DOM. In this case (and for ChildNode), it was easier not to wrap the existing versions of these functions (built-in or polyfilled) since everything goes through Shady DOM's implementation of `insertBefore` and `removeChild` anyway, so Shady DOM doesn't actually use them. However, without them, we won't have support for these features without enabling Shady DOM.

packages/tests/webcomponentsjs_/parent-node/prepend.html
```diff
@@ -0,0 +1,95 @@
...
+<html>
+<head>
+  <title>ParentNode prepend</title>
+  <script>
+    for (const item of [
+      Document.prototype,
+      DocumentFragment.prototype,
+      Element.prototype,
+    ]) {
+      delete item.prepend;
```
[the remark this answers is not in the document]
bicknellr: (replied in earlier comment)

packages/shadydom/src/patch-native.js
sorvell: I think the concern was [IE/Edge](https://github.com/webcomponents/polyfills/blob/4c539009f045723cee214f5212a5e42b2601f5f6/packages/shadydom/src/patch-native.js#L247).

It would be good to verify (perhaps just via one-off manual testing rather than in the code itself) that these properties do not exist *higher* on the prototype chain than where we patch them on any browser. This would cause the patch to fail.

packages/shadydom/src/utils.js
```diff
@@ -230,3 +230,17 @@ export const assign = (target, source) => {
...
+    return (typeof args[0] === 'string' ? document.createTextNode(args[0]) : args[0]);
```
sorvell, line 238: Let's factor this out since it's done 2x?

packages/tests/webcomponentsjs_/parent-node/append.html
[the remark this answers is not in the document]
sorvell: OK, I missed the fact that this isn't a shadydom test.

packages/webcomponentsjs/ts_src/platform/parent-node/append.ts
sorvell: I'm just not sure this is worth doing since there is no browser that has Shadow DOM but does not have ParentNode/ChildNode.

packages/shadydom/src/patch-native.js
bicknellr: I checked the earliest version of the browsers implementing `append` (Edge 17, Firefox 49, Chrome 54, Safari 10) and they all have the descriptor on `Document`, `DocumentFragment`, and `Element` prototypes but not the `HTMLDocument`, `HTMLElement`, or `ShadowRoot` (if it exists) prototypes, so they should be alright assuming they haven't regressed and moved them around since.

packages/webcomponentsjs/ts_src/platform/parent-node/append.ts
sorvell: I didn't realize that there was any browser that had this configuration. It seems ok then.

sorvell (approved):

packages/shadydom/src/utils.js
```diff
@@ -230,3 +230,35 @@ export const assign = (target, source) => {
...
 };
+
+/**
+ * Converts a single value to a node for `convertNodesIntoANode`.
+ *
+ * @param {*} arg
+ * @return {!Node}
+ */
+const convertIntoANode = (arg) => {
+  return !(arg instanceof Node) ? document.createTextNode(String(arg)) : arg;
```
bicknellr: FYI, I removed this explicit conversion to string because browsers don't seem to do this and will throw when trying to `append` a symbol.

PR #393, opened 2020-10-12
Polyfill `Element#getAttributeNames`.

bicknellr:
Adds a polyfill for [`Element#getAttributeNames`](https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttributeNames).

### DIFF
--- a/packages/webcomponentsjs/ts_src/platform/get-attribute-names.ts
+++ b/packages/webcomponentsjs/ts_src/platform/get-attribute-names.ts
@@ -1,0 +1,20 @@
+/**
+@license
+Copyright (c) 2020 The Polymer Project Authors. All rights reserved.
+This code may only be used under the BSD style license found at
+http://polymer.github.io/LICENSE.txt The complete set of authors may be found at
+http://polymer.github.io/AUTHORS.txt The complete set of contributors may be
+found at http://polymer.github.io/CONTRIBUTORS.txt Code distributed by Google as
+part of the polymer project is also subject to an additional IP rights grant
+found at http://polymer.github.io/PATENTS.txt
+*/
+
+const Element_prototype = Element.prototype;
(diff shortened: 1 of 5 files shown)


bicknellr:
@justinfagnani, isn't there some attribute ordering problem with IE11? Does this polyfill need to address that?

justinfagnani:
> @justinfagnani, isn't there some attribute ordering problem with IE11? Does this polyfill need to address that?

There's no way to address this since the ordering is lost during parsing. We're already robust to ordering changes when using `.attributes` though, so we're ok here.

packages/webcomponentsjs/ts_src/platform/get-attribute-names.ts
```diff
@@ -0,0 +1,29 @@
...
+    Object.getOwnPropertyDescriptor(Node.prototype, 'attributes');
+// In Safari 9, the `attributes` descriptor's getter is undefined. In Chrome 41,
+// the `attributes` descriptor is a data descriptor on each Element instance.
+const getAttributes = attributesDescriptor?.get ??
+    function(this: Element) { return this.attributes; };
+const map = Array.prototype.map;
+
+if (!Element_prototype.hasOwnProperty('getAttributeNames')) {
+  Element_prototype.getAttributeNames = function getAttributeNames(this: Element): Array<string> {
+    return map.call(getAttributes.call(this), attr => attr.name) as Array<string>;
```
justinfagnani, line 27: Instead of storing the `.attributes` getter, why not just call `this.attributes` here?

bicknellr: It's just so that this won't call into anything else that might wrap the `attributes` getter after this polyfill is applied.

justinfagnani (approved):

PR #394, opened 2020-10-12
Polyfill `Object.is`.

bicknellr:
Adds a polyfill for [`Object.is`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is).


sorvell (approved):

justinfagnani:
The code here looks fine, I just want to clarify that we do intend to add more JS-only polyfills rather than relying on core-js. IE11 is the only browser we support that needs this, and I thought we might rely on core-js for other JS polyfills.

bicknellr:
(relaying from other chat) We're going to ask users to separately include an `Object.is` polyfill if needed.

PR #398, opened 2020-10-17
Improve the ESLint setup

abdonrd:
I do the same work as https://github.com/Polymer/lit-html/pull/1351, but just with the ESLint for now to make it easier to review.
Later in other PR I can add the Prettier and the husky & lint-staged setups.

---

- Move the `lint` command to the root.
  - Now there is one entry point to lint all the project.
  - Having lerna in between, made us lose the colors and links in the terminal that ESLint provides. Now we have them again; a nice improvement to the development experience!

---

- Now we also use the `.gitignore` file in ESLint.

---

- Now we have the `.gitignore` files unified.

---

- Now all the project is linted.
  - In new PRs we can try to remove temporally ignored rules, inlined ignored comments, etc.

---


packages/tests/html-imports/html/es-module.html
```diff
@@ -50,8 +50,8 @@
 
         try {
           document.head.appendChild(link);
+        // eslint-disable-next-line no-empty
         } catch (e) {
```
bicknellr, line 54:
```suggestion
```

(Also depends on polyserve as mentioned in an earlier comment.)

packages/tests/shadydom/active-element.html
```diff
@@ -481,6 +481,7 @@
     r_1_1.focus();
     try {
       document.activeElement = "abc";
+    // eslint-disable-next-line no-empty
     } catch(e) {}
```
bicknellr, line 485:
```suggestion
    }
```

(also depends on polyserve)

packages/tests/shadydom/active-element.html
```diff
@@ -496,6 +497,7 @@
     r_1_1.focus();
     try {
       ShadyDOM.wrapIfNeeded(ShadyDOM.wrapIfNeeded(r_1).shadowRoot).activeElement = "abc";
+    // eslint-disable-next-line no-empty
     } catch(e) {}
```
bicknellr, line 501:
```suggestion
    }
```

(also depends on polyserve)

packages/tests/shadydom/shady-content.html
```diff
@@ -1068,8 +1068,8 @@
       var div = document.createElement('div');
       ShadyDOM.wrapIfNeeded(div).innerHTML = '<x-dist><div></div></x-dist>';
       var h1 = ShadyDOM.wrapIfNeeded(div).firstChild;
-      var h2 = createPatchedDocumentFragment();;
-      ShadyDOM.wrapIfNeeded(document.body).appendChild(h1);
+      var h2 = createPatchedDocumentFragment();
+    ShadyDOM.wrapIfNeeded(document.body).appendChild(h1);
```
bicknellr, line 1072:
```suggestion
      ShadyDOM.wrapIfNeeded(document.body).appendChild(h1);
```

packages/tests/shadydom/shady-dynamic.html
```diff
@@ -1171,6 +1171,7 @@
     assert.deepEqual(composed, [fallback, ref]);
     ShadyDOM.wrapIfNeeded(slot).removeChild(fallback);
     ShadyDOM.flush();
+    // eslint-disable-next-line no-redeclare
     var composed = Array.from(ShadyDOM.nativeTree.childNodes(e));
```
bicknellr, line 1175:
```suggestion
    composed = Array.from(ShadyDOM.nativeTree.childNodes(e));
```

packages/url/url.js
```diff
@@ -14,6 +14,7 @@ Window.prototype.forceJURL = false;
       var u = new URL('b', 'http://a');
       u.pathname = 'c%20d';
       hasWorkingUrl = u.href === 'http://a/c%20d';
+    // eslint-disable-next-line no-empty
     } catch(e) {}
```
bicknellr, line 18:
```suggestion
    }
```

(depends on transpiler)

packages/url/urltestparser.js
```diff
@@ -1,6 +1,7 @@
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
 function URLTestParser(input) {
   var relativeSchemes = ["ftp", "file", "gopher", "http", "https", "ws", "wss"],
-      tokenMap = { "\\": "\\", n: "\n", r: "\r", s: " ", t: "\t", f: "\f" }
+      tokenMap = { "\\": "\\", n: "\n", r: "\r", s: " ", t: "\t", f: "\f" },
```
bicknellr, line 4: 😱  If only the community / TC39 could somehow figure out how to make it OK to make a breaking change to ES so ASI could be removed.

packages/webcomponentsjs/ts_src/platform/child-node/after.ts
```diff
@@ -10,15 +10,16 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
 export {};
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
 type Constructor<T> = new (...args: Array<any>) => T;
 
 const nativeInsertBefore = Node.prototype.insertBefore;
 const nativeGetParentNode =
-    Object.getOwnPropertyDescriptor(Node.prototype, 'parentNode')?.get! ??
+    Object.getOwnPropertyDescriptor(Node.prototype, 'parentNode')?.get ??
```
bicknellr, line 18: Do these actually compile without the `!`s? I'm pretty sure I had to put them in because TS knows it's not guaranteed that some arbitrary descriptor will have a `get`.

packages/tests/html-imports/html/imports/es-module-1.js
```diff
@@ -1,3 +1,4 @@
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
 import {message} from './es-module-2.js';
```
bicknellr, line 2:
```suggestion
import './es-module-2.js';
```

I don't *think* this was relevant to the test.

bicknellr (commented):
Thanks for doing this!

There seems to be something going on with the `Object.assign` polyfill, which is weird given that it wasn't modified here.

packages/tests/custom-elements/html/polyfill-define-lazy.html
```diff
@@ -50,13 +50,13 @@
 
     test('name must contain a dash', function() {
       assert.throws(function () {
-        customElements.polyfillDefineLazy('xfoo', () => {prototype: Object.create(HTMLElement.prototype)});
+        customElements.polyfillDefineLazy('xfoo', () => {Object.create(HTMLElement.prototype)});
```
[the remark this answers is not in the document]
bicknellr: I hope this actually works, I don't think the intention here would have been anything else.

packages/html-imports/src/html-imports.js
```diff
@@ -117,6 +117,7 @@
           const u = new URL('b', 'http://a');
           u.pathname = 'c%20d';
           Path.__workingURL = (u.href === 'http://a/c%20d');
+        // eslint-disable-next-line no-empty
         } catch (e) {}
```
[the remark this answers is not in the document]
abdonrd: Ok! But we need the `}` to close the `try`.

```suggestion
        }
```

packages/webcomponentsjs/ts_src/platform/child-node/after.ts
abdonrd: If you want the `!`, then you need to remove the previous `?`.
Because can't ensure a not-null after a posible undefined value.

bicknellr (commented):
Sorry about my misleading comments earlier! I thought `try` didn't require a `catch` or `finally` block anymore; apparently it's just the catch binding.

Also, I found out why the `Object.assign` tests are breaking: the `export {};` in the original source of `packages/webcomponentsjs/ts_src/platform/es6-misc.ts` is now being included in the output as a result of TS 4.0 . This file is referenced in the `Object.assign` tests as a classic script, so it breaks on the export syntax. Changing the path `../node_modules/@webcomponents/webcomponentsjs/src/platform/es6-misc.js` to `../node_modules/@webcomponents/webcomponentsjs/bundles/webcomponents-pf_js.js` should fix this.

abdonrd:
Thanks @bicknellr! Fixed!

packages/webcomponentsjs/ts_src/platform/child-node/after.ts
bicknellr: Oh interesting, I had always assumed the `!` was applying to specifically the `.get` part, but it sounds like it actually applies to the entire expression leading up to that point.

abdonrd: Exactly!

abdonrd:
@bicknellr I just merged with `master` to resolve conflicts!

The CLA is red again 😕

bicknellr:
Thanks again!

PR #399, opened 2020-10-19
test: add unit test to custom-elements-scoped

manolakis:
I just added a set of unit test. They are using web test runner and playwright so they are executed on latest versions of Chromium, Webkit and Firefox.

To run them just execute `npm run test:wtr`. In order to debug you can run `npm run test:wtr -- --watch`

Some of them are currently failing:
- global registry: seems that with the polyfill the user can define the same element class twice with different tag names. Also, the upgrade method seems to be undefined.
- scoped registry: have the same problems than the global registry.
- ShadowRoot: `importNode` is throwing an Error in some situations (`Element is being upgrade outside of a valid tree scope`) and tries to access to the prototype of an undefined element in others.
- Element: `innerHTML` seems to fail in some situations.

Hope this helps :)

/cc @kevinpschaaf


LarsDenBakker:
Ping @kevinpschaaf

PR #400, opened 2020-10-19
Polyfill `Element#matches`.

### LINKED ISSUE #397: Add polyfill for `Element.matches`
## Description

Older browsers have prefixed versions, see: https://developer.mozilla.org/en-US/docs/Web/API/Element/matches

## Motivation

It's used in LitElement's queryAssignedNodes.

bicknellr:
Fixes #397. As mentioned in the bug, `Element#matches` exists with a vendor-prefixed in the browsers that we polyfill, so this just copies the function over using the standard name.

### DIFF
--- a/packages/webcomponentsjs/ts_src/platform/matches.ts
+++ b/packages/webcomponentsjs/ts_src/platform/matches.ts
@@ -1,0 +1,20 @@
+/**
+@license
+Copyright (c) 2020 The Polymer Project Authors. All rights reserved.
+This code may only be used under the BSD style license found at
+http://polymer.github.io/LICENSE.txt The complete set of authors may be found at
+http://polymer.github.io/AUTHORS.txt The complete set of contributors may be
+found at http://polymer.github.io/CONTRIBUTORS.txt Code distributed by Google as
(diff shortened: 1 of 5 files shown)


packages/webcomponentsjs/ts_src/platform/matches.ts
```diff
@@ -0,0 +1,20 @@
...
+found at http://polymer.github.io/PATENTS.txt
+*/
+
+export {};
+
+const Element_prototype = Element.prototype;
+
+if (!Element_prototype.hasOwnProperty('matches')) {
+  Element_prototype.matches =
+      ((Element_prototype as any).webkitMatchesSelector as Element['matches']) ??
```
sorvell, line 18: Maybe add `mozMatches` (but not sure it's relevant at all)?

sorvell (approved):

packages/webcomponentsjs/ts_src/platform/matches.ts
bicknellr: Firefox 34 has the unprefixed version. I think the `webkit...` one might even be unnecessary here, the MDN page claims that these two should cover IE9+.

PR #412, opened 2020-12-12
test: make test intention clearer

manolakis:
@kevinpschaaf I did some changes in the test to make it more clear.


manolakis:
@kevinpschaaf I just found an issue when some components are created inside a document-fragment, so I added some test to reproduce it.

kevinpschaaf (approved):
LGTM

PR #413, opened 2021-01-07
DO NOT MERGE: Testing if Safari 11 on Sauce is currently broken

bicknellr:
Safari 11 on Sauce via the Polymer CLI is failing on other repos, but it seems like the result of bad WebDriver commands being sent to the browser rather than an actual code problem. The last tests here passed on Safari 11 on Sauce, so I'm pushing a random change to see if Safari 11 starts failing here also.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+random change
+
 # Web Components Polyfills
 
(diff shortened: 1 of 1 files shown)


bicknellr:
Hmm, Safari 11 seems to work here. Maybe this is a Polymer CLI vs WCT issue?

PR #416, opened 2021-01-13
When `EventTarget.prototype` exists but is not in the prototype chain of `window`, polyfill installation breaks.

bicknellr:
Firefox 33 has `EventTarget`, but `EventTarget.prototype` is not in the prototype chain of `window`. Currently, if `EventTarget` exists, then the `__shady_native_`-prefixed functions are only added to `EventTarget.prototype`. In Firefox 33, this causes the polyfills to throw when a call to one of these functions on `window` is attempted later during startup.

Tested as http://cl/351475102 .

### DIFF
--- a/packages/shadydom/src/patch-native.js
+++ b/packages/shadydom/src/patch-native.js
@@ -110,6 +110,13 @@ export const addNativePrefixedProperties = () => {
   ];
   if (window.EventTarget) {
     copyProperties(window.EventTarget.prototype, eventProps);
+
+    // In Firefox 33, `EventTarget` exists, but `EventTarget.prototype` is not
+    // in the prototype chain of `window` and, strangely,
+    // `window instanceof EventTarget` still returns true.
+    if (window[NATIVE_PREFIX + 'addEventListener'] === undefined) {
+      copyProperties(Window.prototype, eventProps);
+    }
   } else {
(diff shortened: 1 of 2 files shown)


packages/shadydom/src/patch-native.js
```diff
@@ -110,6 +110,13 @@ export const addNativePrefixedProperties = () => {
   ];
   if (window.EventTarget) {
     copyProperties(window.EventTarget.prototype, eventProps);
+
+    // In Firefox 33, `EventTarget` exists, but `EventTarget.prototype` is not
+    // in the prototype chain of `window` and, strangely,
+    // `window instanceof EventTarget` still returns true.
+    if (window[NATIVE_PREFIX + 'addEventListener'] === undefined) {
```
kevinpschaaf, line 117: Does `EventTarget` exist in `Node.prototype`? (the else clause also deals with patching Node up)

bicknellr: Yes, `Node` seems to be set up correctly:

Screen Shot 2021-01-13 at 14 44 33

kevinpschaaf: Ok, thanks for checking.

kevinpschaaf (approved):
LGTM

PR #433, opened 2021-02-19
Upstream cl/345169174: Fix calls to XMLHttpRequest#getResponseHeader to accept null as a return value.

bicknellr:
This PR upstreams http://cl/345169174 , which fixes a warning caused by this update to the Closure externs: https://github.com/google/closure-compiler/commit/113514db81ef44e2c145f9fc76fa502cd57010ab .


bicknellr:
WCT's server doesn't support `??`, which makes it throw while serving to older browsers.

bicknellr:
One of the HTML Imports test files was referencing the source directly, rather than the compiled build output.

aomarks (approved):

PR #439, opened 2021-04-25
feat(scoped-registry): applies the polyfill only once

### LINKED ISSUE #438: [scoped-custom-element-registry] should always only be applied once
### Description

Right now if you load the polyfill multiple times from different locations. (e.g. nested node_modules) it would override each other.

Adding something like this

```js
if (!window.__scopedCustomElementRegistryLoadedPolyfill) {
  window.__scopedCustomElementRegistryLoadedPolyfill = true;
```

at the top should be good enough to make sure the polyfill does not fail in these cases.

What do you think?

manolakis:
Checks if `ShadowRoot` prototype supports the `createElement` method to determine if the polyfill should be applied or not.

Fixes #438

@kevinpschaaf


kevinpschaaf (approved):
LGTM

PR #441, opened 2021-04-26
fix(scoped-registry): native HTMLElement

### LINKED ISSUE #440: [scoped-custom-element-registry] throws error trying to define a CE extending the native HTMLElement
### Description

The polyfill throws an Error indicating that there is an illegal constructor if we try to define a CE which class has been created before applying the polyfill, so it's using the native HTMLElement function instead of the patched one.

### Steps to reproduce

```js
class CustomElementClass extends HTMLElement {}

await import('scoped-custom-element-registry.min.js');

customElements.define('my-example', CustomElementClass); // throws Error
```

#### Expected behavior

Define the component.

#### Actual behavior

Throws Error

manolakis:
This commit checks the CE class hierarchy and applies a patch on it in case the native HTMLElement is used by the CE to be defined.

Fixes #440 

@kevinpschaaf

### DIFF
--- a/packages/scoped-custom-element-registry/CHANGELOG.md
+++ b/packages/scoped-custom-element-registry/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 <!-- ## Unreleased -->
 
+- Fix to allow definition of custom elements whose classes have been created before applying the polyfill.
 - Run `connectedCallback` on upgraded elements. Fixes #442.
 - Checks if ShadowRoot prototype supports the createElement method to determine if the polyfill should be applied or not.
 
--- a/packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js
+++ b/packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js
@@ -307,12 +307,31 @@ if (!ShadowRoot.prototype.createElement) {
     }
   };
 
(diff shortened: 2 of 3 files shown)


kevinpschaaf (commented):
My only feedback here is it seems unfortunate to add this tax to all definitions for the case that the polyfill is loaded out of order of definitions. As a general rule, polyfills that simulate platform features need to be loaded before using the feature, so I’m sorta curious how this comes up in practice?

Suggestions below for maybe reducing the tax, wdyt?

daKmoR:
As long as there is no browser implementation we would like to treat it more as an implementation detail than a polyfill in `ScopedElements`. 

We already have a LOT of users doing 

```js
class MyElement extends ScopedElementsMixin(LitElement) {}
```

and it just works, no need to load a polyfill or anything. We would like to keep that until the spec and the polyfill are more matured. We sort of want to avoid letting/forcing the user to load the polyfill manually as that can lead to problems like 
- polyfill/implementation mismatch
- loading of multiple polyfills
- needing to load this polyfill in every test file you write
- ...

For that reason, we want to load the polyfill within the Mixin so it "just works".

manolakis:
@kevinpschaaf I just removed the tax to all definitions, so now is going to be applied __only__ for those elements created before applying the polyfill. If you prefer to not having this change in the polyfill feel free to close the PR.

Thanks!

kevinpschaaf (approved):
Seems like an ok compromise. Thanks!

PR #446, opened 2021-05-12
[scoped-custom-element-registry] Run `connectedCallback` on upgraded elements

### LINKED ISSUE #442: [scoped-custom-element-registry] lazy definitions seems to not upgrade the element if the same tag name has been defined in global registry
### Description

if we do a lazy component definition over a scoped registry using a tag name previously defined in the global registry, the components are not upgraded.

@kevinpschaaf 

### Example

See https://github.com/lazka/scoped-test-v2/blob/main/demo.js for a minimal example.

The `SomethingDifferent` component isn't visible if it is registered globally first and defineScopedElement() is called after the first update.

Two things:

Removing the first `customElements.define` fixes things
Calling `this.defineScopedElement` earlier (before render) fixes things as well

### Steps to reproduce

build the previous example and open the HTML inside the dist folder.

#### Expected behavior

It should upgrade the component

#### Actual behavior

The component is not upgraded

### Version

scoped-custom-element-registry v0.0.1

kevinpschaaf:
In the "upgrade" case (when an element is registered and there are elements in the `_awaitingUpgrade` map for that tag name), we were failing to run `connectedCallback()`. If the stand-in element definition was defined in the global registry first, then we would have no-opted the initial `connectedCallback()`, and so when manually upgrading the element later, we need to be sure to run its `connectedCallback` if `isConnected === true`.

Fixes #442


packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js
```diff
@@ -88,7 +88,7 @@ if (!ShadowRoot.prototype.createElement) {
         this._awaitingUpgrade.delete(tagName);
         for (const element of awaiting) {
           pendingRegistryForElement.delete(element);
-          upgrade(element, definition);
```
justinfagnani, line 91: is there test coverage for this change, or were the tests already failing?

kevinpschaaf: Oh weird. I added a test... not sure what happened to it.

kevinpschaaf: Ok, added the test. PTAL.

.github/workflows/tests.yml
```diff
@@ -42,6 +42,27 @@ jobs:
       - name: WCT local (Chrome/Firefox)
         run: npm test
 
+  wtr-local:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
```
justinfagnani, line 50: can we use `actions/setup-node@v2` and set `node-version: 14`?

web-test-runner.config.js
```diff
@@ -6,26 +6,17 @@ const defaultBrowsers = [
   playwrightLauncher({product: 'firefox', concurrency: 1}),
 ];
 
-const envBrowsers = process.env.BROWSERS?.split(',').map((product) =>
-  playwrightLauncher({product})
-);
+const envBrowsers =
```
justinfagnani, line 9: If we use Node 14 we don't need these changes

packages/scoped-custom-element-registry/test/scoped-registry.test.html.js
```diff
@@ -54,5 +54,42 @@ describe('Scoped Registry', () => {
 
       expect(() => new CustomElementClass()).to.throw();
     });
+
+    it('should allow defining in global registry and scoped registery', () => {
+      const container = document.createElement('div');
+      document.body.appendChild(container);
```
justinfagnani, line 60: remove container at end of test

packages/scoped-custom-element-registry/test/scoped-registry.test.html.js
```diff
@@ -54,5 +54,42 @@ describe('Scoped Registry', () => {
...
+      container.innerHTML = `<${tagName} attr="doc"></${tagName}`;
+      const docEl = container.firstChild;
+      expect(docEl).to.be.instanceOf(CustomElementClass);
+      expect(docEl.connected).to.be.true;
+      expect(docEl.attributeChanged).to.equal('attr:doc');
+
+      // Create scoped element, then upgrade via scoped registry
+      const registry = new CustomElementRegistry();
+      const shadowRoot = getShadowRoot(registry);
+      document.body.appendChild(shadowRoot.host);
```
justinfagnani, line 86: remove at end of test

justinfagnani (approved):

packages/scoped-custom-element-registry/test/scoped-registry.test.html.js
kevinpschaaf: Done

packages/scoped-custom-element-registry/test/scoped-registry.test.html.js
kevinpschaaf: Done

web-test-runner.config.js
kevinpschaaf: Will test whether it just works, didn't want to get sucked into a big task of debugging random infra stuff.

.github/workflows/tests.yml
kevinpschaaf: Will test whether it just works, didn't want to get sucked into a big task of debugging random infra stuff.

PR #447, opened 2021-05-13
[scoped-custom-element-registry] Resolve whenDefined promise if already defined.

### LINKED ISSUE #443: [scoped-custom-element-registry]: whenDefined() doesn't work when already defined
```html
<script type="module">
import 'https://jspm.dev/@webcomponents/scoped-custom-element-registry/scoped-custom-element-registry.min.js';

class SomethingDifferent extends HTMLElement {}

customElements.define('something-different', SomethingDifferent);

customElements.whenDefined('something-different').then(() => {
    console.log("works"); // this doesn't get called
});
</script>
```

* whenDefined promise doesn't do anything if the element was defined before
* without the polyfill this works as documented

### LINKED ISSUE #444: [scoped-custom-element-registry]: insertAdjacentHTML throws exception
Exception is thrown here:
https://github.com/webcomponents/polyfills/blob/041aff1a2408a5854c880e45dfad150556e51bcb/packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js#L345

Reason: native `insertAdjacentHTML` returns `undefined`, but `WeakMap`s only support objects as keys:
- https://w3c.github.io/DOM-Parsing/#extensions-to-the-element-interface
- step 4 in https://tc39.es/ecma262/#sec-weakmap.prototype.set

Can be reproduced with the following example, which should spell out "success" (I can get the example to work by commenting out the line `installScopedCreationMethod(Element, 'insertAdjacentHTML');`):
```html
<span>u</span><span id="host"></span>
<script>
    class MyWidgetGlobal extends HTMLElement {
        constructor() {
            super()
            this.attachShadow({ mode: 'open' })
            this.shadowRoot.innerHTML = '<span>s</span>'
        }
    }
    class MyWidgetScoped extends HTMLElement {
        constructor() {
            super()
            this.attachShadow({ mode: 'open' })
            this.shadowRoot.innerHTML = '<span>c</span>'
        }
    }
    window.customElements.define('my-widget', MyWidgetGlobal)
    try {
        const span = document.querySelector('span')
        span.insertAdjacentHTML('beforebegin', '<my-widget></my-widget>')
        // next line won't execute
        span.insertAdjacentText('afterend', 'c')
    } catch (ex) {
        console.log(ex)
    }
    const scope = new CustomElementRegistry()
    scope.define('my-widget', MyWidgetScoped)
    const host = document.querySelector('#host')
    const shadowRoot = host.attachShadow({ mode: 'open', customElements: scope })
    shadowRoot.innerHTML = '<span>es</span>'
    try {
        const span = shadowRoot.querySelector('span')
        span.insertAdjacentHTML('beforebegin', '<my-widget></my-widget>')
        // next line won't execute
        span.insertAdjacentText('afterend', 's')
    } catch (ex) {
        console.log(ex)
    }
</script>
```

kevinpschaaf:
Also ensure the promise resolves to the constructor, per spec.

Fixes #443

### DIFF
--- a/packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js
+++ b/packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js
@@ -94,7 +94,7 @@ if (!ShadowRoot.prototype.createElement) {
       // Flush whenDefined callbacks
       const resolver = this._definedResolvers.get(tagName);
       if (resolver) {
-        resolver();
+        resolver(elementClass);
       }
       return elementClass;
     }
@@ -121,6 +121,10 @@ if (!ShadowRoot.prototype.createElement) {
         promise = new Promise((r) => (resolve = r));
         this._definedPromises.set(tagName, promise);
         this._definedResolvers.set(tagName, resolve);
+        const definition = this._getDefinition(tagName);
+        if (definition !== undefined) {
+          resolve(definition.elementClass);
+        }
       }
       return promise;
     }
--- a/packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js
+++ b/packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js
@@ -34,8 +34,7 @@ if (!ShadowRoot.prototype.createElement) {
     constructor() {
       this._definitionsByTag = new Map();
       this._definitionsByClass = new Map();
-      this._definedPromises = new Map();
-      this._definedResolvers = new Map();
+      this._whenDefinedPromises = new Map();
       this._awaitingUpgrade = new Map();
     }
 
@@ -92,9 +91,10 @@ if (!ShadowRoot.prototype.createElement) {
         }
       }
       // Flush whenDefined callbacks
-      const resolver = this._definedResolvers.get(tagName);
-      if (resolver) {
-        resolver(elementClass);
+      const info = this._whenDefinedPromises.get(tagName);
+      if (info !== undefined) {
+        info.resolve(elementClass);
+        this._whenDefinedPromises.delete(tagName);
       }
       return elementClass;
     }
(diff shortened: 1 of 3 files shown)


packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js
```diff
@@ -94,7 +94,7 @@ if (!ShadowRoot.prototype.createElement) {
       // Flush whenDefined callbacks
       const resolver = this._definedResolvers.get(tagName);
       if (resolver) {
-        resolver();
+        resolver(elementClass);
```
justinfagnani, line 97: I think ideally we should clear the `_definedResolvers` and `_definedPromises` maps. After resolving, anyone who newly calls `whenDefined()` can get a fresh Promise.

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js
```diff
@@ -121,6 +121,10 @@ if (!ShadowRoot.prototype.createElement) {
         promise = new Promise((r) => (resolve = r));
         this._definedPromises.set(tagName, promise);
         this._definedResolvers.set(tagName, resolve);
+        const definition = this._getDefinition(tagName);
```
justinfagnani, line 124: I think this check should go first: If the tag is defined, return `Promise.resolve(definition.elementClass)`, else store a Promise/resolver pair. Then we only have a single place to resolve the promise, and do any additional cleanup like in my other comment.

Nearby cleanup: think it'd be slightly nicer to have one map with an object value than two maps with the same key.

kevinpschaaf: I was assuming whenDefined should return the same promise for all calls with the same tagName; I just checked the [spec](https://html.spec.whatwg.org/multipage/custom-elements.html#dom-customelementregistry-whendefined) and it should but only until the element is defined, after that they get a fresh resolved promise. So I think that fits with what you suggested.

Made that change.

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js
kevinpschaaf: Done

justinfagnani (approved):

PR #452, opened 2021-06-15
Convert all multi-line import statements to single line

bicknellr:
There's an internal tool that breaks on multiline imports. It's being actively replaced, but for now we need to work around it.

This PR is part of http://cl/379640941, which is waiting on internal tests to finish. I also tested that CL + #455 last night and only saw 5 failures, which were fixed locally after reverting changes from #455.


aomarks (approved):

bicknellr:
Internal tests are clean. See comments in http://cl/379640941.

PR #457, opened 2021-06-30
Add TS externs.

bicknellr:
This PR adds externs for released packages in TypeScript's `.d.ts` format.

The original JSCompiler types from which these were converted conflate types that are used only within the package to describe internal objects or expand known built-in types for use by the package itself with types that are intended for the user of the package. The new types only expose the package's interface in the externs file and keep those intended for internal use in an `env.d.ts` file in each package's source folder.

The TS externs are included in the published packages next to their JSCompiler counterparts, but are not set as the `types` field because they aren't relevant to some of the packages' `main` module (i.e. Shady CSS).

This PR also generates a new `webcomponents-bundle.d.ts` file next to `webcomponents-bundle.js`, which bundles all of the individual polyfills' externs.


packages/custom-elements/externs/custom-elements.d.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * @externs
```
rictic, line 2: here and elsewhere: I don't think this file should be marked as `@externs`. In the internal bazel rules, a ts_declaration() automatically produces externs

packages/custom-elements/ts_src/env.d.ts
```diff
@@ -0,0 +1,84 @@
+/**
+ * @externs
```
rictic, line 2: ditto, re `@externs`

packages/html-imports/externs/html-imports.d.ts
```diff
@@ -0,0 +1,57 @@
+/**
+ * @externs
```
rictic, line 2: `@externs`

packages/shadycss/externs/shadycss.d.ts
```diff
@@ -0,0 +1,71 @@
+/**
+ * @externs
```
rictic, line 2: `@externs`

packages/shadycss/src/env.d.ts
```diff
@@ -0,0 +1,31 @@
+/**
+ * @externs
```
rictic, line 2: `@externs`

packages/shadydom/externs/shadydom.d.ts
```diff
@@ -0,0 +1,37 @@
+/**
+ * @externs
```
rictic, line 2: `@externs`

packages/shadydom/src/env.d.ts
```diff
@@ -0,0 +1,60 @@
+/**
+ * @externs
```
rictic, line 2: `@externs`

packages/template/externs/template.d.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * @externs
```
rictic, line 2: `@externs`

packages/custom-elements/externs/custom-elements.d.ts
```diff
@@ -0,0 +1,23 @@
...
+ * Copyright (c) 2021 The Polymer Project Authors. All rights reserved. This
+ * code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt The complete set of authors may be found
+ * at http://polymer.github.io/AUTHORS.txt The complete set of contributors may
+ * be found at http://polymer.github.io/CONTRIBUTORS.txt Code distributed by
+ * Google as part of the polymer project is also subject to an additional IP
+ * rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+
+// This file needs to be a module for the declarations to apply globally.
```
rictic, line 13: here and elsewhere: kinda the opposite is the case. declarations in non-modules are global by default

but for consistency and our own sanity, we prefer all files to be modules

packages/webcomponentsjs/externs/webcomponents.d.ts
```diff
@@ -0,0 +1,19 @@
+/**
+ * @externs
```
rictic, line 2: `@externs`

rictic (approved):
Just some documentation / clarity comments. Please do address, but otherwise LGTM

packages/custom-elements/externs/custom-elements.d.ts
bicknellr: Yeah, this comment doesn't make sense; updated.

packages/custom-elements/externs/custom-elements.d.ts
bicknellr: Removed.

packages/custom-elements/ts_src/env.d.ts
bicknellr: removed

packages/html-imports/externs/html-imports.d.ts
bicknellr: removed

packages/shadycss/externs/shadycss.d.ts
bicknellr: removed

packages/shadycss/src/env.d.ts
bicknellr: removed

packages/shadydom/externs/shadydom.d.ts
bicknellr: removed

packages/shadydom/src/env.d.ts
bicknellr: removed

packages/template/externs/template.d.ts
bicknellr: removed

packages/webcomponentsjs/externs/webcomponents.d.ts
bicknellr: removed

PR #463, opened 2021-07-28
[ShadyDOM] Adds `element(s)FromPoint` to document and shadowRoot

### LINKED ISSUE #69: [ShadyDOM] shadowRoot.elementFromPoint is undefined when using Selenium and Firefox
Running a Selenium test on Firefox and trying to click on an element inside a shadow root, causes the click command to fail with

`WebDriverException: e.elementFromPoint is not a function`

This originates in https://github.com/SeleniumHQ/selenium/blob/77d2ac91eb7c709275b67d9dc4fd0563a2aaab64/javascript/firefox-driver/js/utils.js#L703 where Selenium assumes that every shadowRoot has an `elementFromPoint` function. As you can see in http://jsbin.com/potatudafa/edit?html,console, a Shady shadowRoot does not have such a function

### LINKED ISSUE #461: [ShadyDOM] Add support for element(s)FromPoint
## Description

We need to add this to be compatible with Selenium. It's also related to quite an old [issue](https://github.com/webcomponents/polyfills/issues/69).

sorvell:
Fixes #461.
Fixes #69.


kevinpschaaf (approved):
LGTM, algorithm makes sense afaict.

kevinpschaaf (changes requested):
It looks like IE/old Edge tests are failing due to `elementsFromPoint` being prefixed to `msElementsFromPoint`: 
* https://www.caniuse.com/?search=elementsFromPoint
* https://docs.microsoft.com/en-us/previous-versions/hh772121(v=vs.85)

kevinpschaaf (approved):
LGTM

PR #469, opened 2021-08-05
Allow event listener options to be specified with a function in addition to an object or boolean

### LINKED ISSUE #468: [ShadyDOM] Event listener options cannot be specified with a function in IE
### Description

ShadyDOM's `addEventListener` [implementation](https://github.com/webcomponents/polyfills/blob/master/packages/shadydom/src/patch-events.js#L45-L64) only allows non-boolean event listener options passed in as objects. If a function is passed, the event will always capture, which is incorrect. Instead of checking if the options are an object, we can probably check that they are not a boolean.

Note, `noPatch` users cannot rely on this patch anyway and need a separate polyfill if they want to use event options.

### Browsers affected

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari
- [X] IE 11

sorvell:
Fixes #468 and #471

* Allows event listener options to be specified with any options that are `instanceof Object`. This allows functions to be used in addition to objects.
* Also fixes `eventPhase` to properly be `Event.AT_TARGET` on retargeted events.

### DIFF
--- a/packages/shadydom/src/patch-events.js
+++ b/packages/shadydom/src/patch-events.js
@@ -193,16 +193,12 @@ function retarget(refNode, path) {
   }
 }
 
-const nativeEventPhase = Object.getOwnPropertyDescriptor(
+const eventPhaseDescriptor = Object.getOwnPropertyDescriptor(
   Event.prototype,
   'eventPhase'
 );
 
 let EventPatches = {
-  get [utils.NATIVE_PREFIX + 'eventPhase']() {
-    return nativeEventPhase.get.call(this);
-  },
-
   get eventPhase() {
     return this.currentTarget === this.target
       ? Event.AT_TARGET
@@ -285,6 +281,12 @@ let EventPatches = {
     this.__propagationStopped = true;
   },
 };
+
+Object.defineProperty(
+  EventPatches,
+  utils.NATIVE_PREFIX + 'eventPhase',
+  eventPhaseDescriptor
+);
 
 function mixinComposedFlag(Base) {
   // NOTE: avoiding use of `class` here so that transpiled output does not
--- a/packages/shadydom/src/patch-events.js
+++ b/packages/shadydom/src/patch-events.js
@@ -514,7 +514,7 @@ export function addEventListener(type, fnOrObj, optionsOrCapture) {
    * @this {HTMLElement}
    * @param {Event} e
    */
-  const wrapperFn = function (e) {
+  const wrapperFn = (e) => {
     // Support `once` option.
     if (once) {
       this[utils.SHADY_PREFIX + 'removeEventListener'](
--- a/packages/shadydom/src/patch-events.js
+++ b/packages/shadydom/src/patch-events.js
@@ -193,7 +193,22 @@ function retarget(refNode, path) {
   }
 }
 
+const nativeEventPhase = Object.getOwnPropertyDescriptor(
+  Event.prototype,
+  'eventPhase'
+);
+
 let EventPatches = {
+  get [utils.NATIVE_PREFIX + 'eventPhase']() {
+    return nativeEventPhase.get.call(this);
+  },
+
+  get eventPhase() {
+    return this.currentTarget === this.target
+      ? Event.AT_TARGET
+      : this[utils.NATIVE_PREFIX + 'eventPhase'];
+  },
+
   /**
    * @this {Event}
    */
--- a/packages/shadydom/src/patch-events.js
+++ b/packages/shadydom/src/patch-events.js
@@ -44,7 +44,10 @@ const supportsEventOptions = (() => {
 
 const parseEventOptions = (optionsOrCapture) => {
   let capture, once, passive, shadyTarget;
-  if (optionsOrCapture && optionsOrCapture instanceof Object) {
+  if (
+    optionsOrCapture &&
+    (typeof optionsOrCapture === 'object' || optionsOrCapture instanceof Object)
+  ) {
     capture = Boolean(optionsOrCapture.capture);
     once = Boolean(optionsOrCapture.once);
     passive = Boolean(optionsOrCapture.passive);
--- a/packages/shadydom/src/patch-events.js
+++ b/packages/shadydom/src/patch-events.js
@@ -271,11 +271,12 @@ let EventPatches = {
   },
 };
 
-const eventPhaseDescriptor = Object.getOwnPropertyDescriptor(
-  Event.prototype,
-  'eventPhase'
-);
-if (eventPhaseDescriptor !== undefined) {
+const hasDescriptors = utils.settings.hasDescriptors;
+
+const eventPhaseDescriptor =
+  hasDescriptors &&
+  Object.getOwnPropertyDescriptor(Event.prototype, 'eventPhase');
+if (eventPhaseDescriptor) {
   Object.defineProperty(EventPatches, 'eventPhase', {
     get() {
       return this.currentTarget === this.target
--- a/packages/shadydom/src/patch-events.js
+++ b/packages/shadydom/src/patch-events.js
@@ -46,7 +46,8 @@ const parseEventOptions = (optionsOrCapture) => {
   let capture, once, shadyTarget;
   if (
     optionsOrCapture &&
-    (typeof optionsOrCapture === 'object' || optionsOrCapture instanceof Object)
+    (typeof optionsOrCapture === 'object' ||
+      typeof optionsOrCapture === 'function')
   ) {
     capture = Boolean(optionsOrCapture.capture);
     once = Boolean(optionsOrCapture.once);
--- a/packages/shadydom/src/patch-events.js
+++ b/packages/shadydom/src/patch-events.js
@@ -45,7 +45,7 @@ const supportsEventOptions = (() => {
 const parseEventOptions = (optionsOrCapture) => {
   let capture, once, shadyTarget;
   if (
-    optionsOrCapture &&
+    optionsOrCapture !== null &&
     (typeof optionsOrCapture === 'object' ||
       typeof optionsOrCapture === 'function')
   ) {
--- a/packages/shadydom/src/patch-events.js
+++ b/packages/shadydom/src/patch-events.js
@@ -193,18 +193,7 @@ function retarget(refNode, path) {
   }
 }
 
-const eventPhaseDescriptor = Object.getOwnPropertyDescriptor(
-  Event.prototype,
-  'eventPhase'
-);
-
 let EventPatches = {
-  get eventPhase() {
-    return this.currentTarget === this.target
-      ? Event.AT_TARGET
-      : this[utils.NATIVE_PREFIX + 'eventPhase'];
-  },
-
   /**
    * @this {Event}
    */
@@ -282,11 +271,27 @@ let EventPatches = {
   },
 };
 
-Object.defineProperty(
-  EventPatches,
-  utils.NATIVE_PREFIX + 'eventPhase',
-  eventPhaseDescriptor
+const eventPhaseDescriptor = Object.getOwnPropertyDescriptor(
+  Event.prototype,
+  'eventPhase'
 );
+if (eventPhaseDescriptor !== undefined) {
+  Object.defineProperty(EventPatches, 'eventPhase', {
+    get() {
+      return this.currentTarget === this.target
+        ? Event.AT_TARGET
+        : this[utils.NATIVE_PREFIX + 'eventPhase'];
+    },
+    enumerable: true,
+    configurable: true,
+  });
+
+  Object.defineProperty(
+    EventPatches,
+    utils.NATIVE_PREFIX + 'eventPhase',
+    eventPhaseDescriptor
+  );
+}
 
 function mixinComposedFlag(Base) {
   // NOTE: avoiding use of `class` here so that transpiled output does not
--- a/packages/shadydom/src/patch-events.js
+++ b/packages/shadydom/src/patch-events.js
@@ -43,25 +43,22 @@ const supportsEventOptions = (() => {
 })();
 
 const parseEventOptions = (optionsOrCapture) => {
-  let capture, once, passive, shadyTarget;
+  let capture, once, shadyTarget;
   if (
     optionsOrCapture &&
     (typeof optionsOrCapture === 'object' || optionsOrCapture instanceof Object)
   ) {
     capture = Boolean(optionsOrCapture.capture);
     once = Boolean(optionsOrCapture.once);
-    passive = Boolean(optionsOrCapture.passive);
     shadyTarget = optionsOrCapture.__shadyTarget;
   } else {
     capture = Boolean(optionsOrCapture);
     once = false;
-    passive = false;
   }
   return {
     shadyTarget,
     capture,
     once,
-    passive,
     nativeEventOptions: supportsEventOptions ? optionsOrCapture : capture,
   };
 };
@@ -382,35 +379,14 @@ function shadyDispatchEvent(e) {
   currentTarget = null;
 }
 
-function listenerSettingsEqual(
-  savedListener,
-  node,
-  type,
-  capture,
-  once,
-  passive
-) {
-  let {
-    node: savedNode,
-    type: savedType,
-    capture: savedCapture,
-    once: savedOnce,
-    passive: savedPassive,
-  } = savedListener;
-  return (
-    node === savedNode &&
-    type === savedType &&
-    capture === savedCapture &&
-    once === savedOnce &&
-    passive === savedPassive
-  );
+function listenerSettingsEqual(savedListener, node, type, capture) {
+  let {node: savedNode, type: savedType, capture: savedCapture} = savedListener;
+  return node === savedNode && type === savedType && capture === savedCapture;
 }
 
-export function findListener(wrappers, node, type, capture, once, passive) {
+export function findListener(wrappers, node, type, capture) {
   for (let i = 0; i < wrappers.length; i++) {
-    if (
-      listenerSettingsEqual(wrappers[i], node, type, capture, once, passive)
-    ) {
+    if (listenerSettingsEqual(wrappers[i], node, type, capture)) {
       return i;
     }
   }
@@ -458,13 +434,9 @@ export function dispatchEvent(event) {
  * @this {EventTarget}
  */
 export function addEventListener(type, fnOrObj, optionsOrCapture) {
-  const {
-    capture,
-    once,
-    passive,
-    shadyTarget,
-    nativeEventOptions,
-  } = parseEventOptions(optionsOrCapture);
+  const {capture, once, shadyTarget, nativeEventOptions} = parseEventOptions(
+    optionsOrCapture
+  );
   if (!fnOrObj) {
     return;
   }
@@ -506,7 +478,7 @@ export function addEventListener(type, fnOrObj, optionsOrCapture) {
     // for few nodes at most, whereas a node will likely have many event listeners).
     // NOTE(valdrin) invoking external functions is costly, inline has better perf.
     // Stop if the wrapper function has already been created.
-    if (findListener(wrappers, target, type, capture, once, passive) > -1) {
+    if (findListener(wrappers, target, type, capture) > -1) {
       return;
     }
   } else {
@@ -613,8 +585,6 @@ export function addEventListener(type, fnOrObj, optionsOrCapture) {
     node: target,
     type: type,
     capture: capture,
-    once: once,
-    passive: passive,
     wrapperFn: wrapperFn,
   });
 
@@ -641,13 +611,9 @@ export function removeEventListener(type, fnOrObj, optionsOrCapture) {
   if (!fnOrObj) {
     return;
   }
-  const {
-    capture,
-    once,
-    passive,
-    shadyTarget,
-    nativeEventOptions,
-  } = parseEventOptions(optionsOrCapture);
+  const {capture, shadyTarget, nativeEventOptions} = parseEventOptions(
+    optionsOrCapture
+  );
   if (unpatchedEvents[type]) {
     return this[utils.NATIVE_PREFIX + 'removeEventListener'](
       type,
@@ -660,7 +626,7 @@ export function removeEventListener(type, fnOrObj, optionsOrCapture) {
   let wrapperFn = undefined;
   let wrappers = getEventWrappers(fnOrObj);
   if (wrappers) {
-    let idx = findListener(wrappers, target, type, capture, once, passive);
+    let idx = findListener(wrappers, target, type, capture);
     if (idx > -1) {
       wrapperFn = wrappers.splice(idx, 1)[0].wrapperFn;
       // Cleanup.
--- a/packages/shadydom/src/patch-events.js
+++ b/packages/shadydom/src/patch-events.js
@@ -272,10 +272,12 @@ let EventPatches = {
   },
 };
 
-const hasDescriptors = utils.settings.hasDescriptors;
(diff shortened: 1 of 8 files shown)


bicknellr (approved):
LGTM

bicknellr:
Really odd edge case. I guess the code where you found this must have been _intentionally_ using a function in this way, since the problem is that `capture` isn't set properly in this case?

kevinpschaaf (approved):
LGTM

packages/shadydom/src/patch-events.js
```diff
@@ -44,7 +44,7 @@ const supportsEventOptions = (() => {
 
 const parseEventOptions = (optionsOrCapture) => {
   let capture, once, passive, shadyTarget;
-  if (optionsOrCapture && typeof optionsOrCapture === 'object') {
+  if (optionsOrCapture && optionsOrCapture instanceof Object) {
```
justinfagnani, line 47: I would do:

```suggestion
  if (optionsOrCapture && (typeof optionsOrCapture === 'object' || typeof optionsOrCapture === 'function')) {
```

because `new Boolean(false) instanceof Object` is true, and to keep the same cross-realm behavior. Plus `typeof` is very fast compared to `instanceof`.

justinfagnani (changes requested):

packages/shadydom/src/patch-events.js
sorvell: `new Boolean(false) == false` so it wouldn't get past the `&&` here at all.

We can put back the `typeof` as a perf optimization.

justinfagnani: I just mean that any `new {Primitive}()` will return an object, `new Boolean(true)` too. That could conceivably flow into options.

packages/webcomponentsjs/ts_src/platform/event-listener-options.ts
```diff
@@ -0,0 +1,85 @@
...
+    origAddEventListener.call(this, type, nativeListener, capture);
+  };
+
+  nativeEventTarget.prototype.removeEventListener = function (
+    type: string,
+    listener: EventListenerOrEventListenerObject | null,
+    options?: boolean | AddEventListenerOptions | undefined
+  ) {
+    const {capture} = parseEventOptions(options);
+    origRemoveEventListener.call(this, type, listener, capture);
```
bicknellr, line 83: This won't remove the correct listener because `listener` here isn't the same listener that's given to `origAddEventListener` above (`nativeListener`). An important thing to note is that keeping a map from the `listener` parameter to the actual listener given to the native function won't be enough to fix this because of how listeners are deduplicated when they're added based on their options.

The 'formdata' event polyfill has to handle a kind of similar situation and has this [`EventListenerArray` class](https://github.com/webcomponents/polyfills/blob/6f17afd6bf42c7383ade9b9819c378c9f504b735/packages/formdata-event/ts_src/event_listener_array.ts) that tracks the deduplication parts. However, the 'formdata' event polyfill only needs to pay attention to listeners for 'formdata' events and isn't actually wrapping these listeners (it only cares when the number of deduplicated 'formdata' event listeners on a target changes between zero and non-zero) so `EventListenerArray` isn't set up to actually provide a full mapping of user-provided listeners to the 'real' listeners; but it would likely be a good starting point for doing that here. There's also a few things in it that aren't needed here: the `push` and `delete` methods are really the important parts.

bicknellr (changes requested):

justinfagnani:
@bicknellr fyi

> Really odd edge case. I guess the code where you found this must have been intentionally using a function in this way, since the problem is that capture isn't set properly in this case?

Yeah, in lit-html we pass the listener from event bindings as both the listener and options arguments. This lets us very easily support an object with `handleEvent` and options, or a method with the `@eventOptions()` decorator - which just sets the options properties on the method.

packages/shadydom/src/patch-events.js
```diff
@@ -44,7 +44,10 @@ const supportsEventOptions = (() => {
 
 const parseEventOptions = (optionsOrCapture) => {
   let capture, once, passive, shadyTarget;
-  if (optionsOrCapture && optionsOrCapture instanceof Object) {
+  if (
+    optionsOrCapture &&
+    (typeof optionsOrCapture === 'object' || optionsOrCapture instanceof Object)
```
justinfagnani, line 49: The only values I know of that are not `typeof o === 'object'` but are `o instanceof Object` are functions. `typeof o === 'function'` should always be faster.

packages/webcomponentsjs/ts_src/platform/event-listener-options.ts
```diff
@@ -38,20 +38,22 @@ if (
...
-    if (optionsOrCapture && optionsOrCapture instanceof Object) {
-      capture = Boolean(optionsOrCapture.capture);
-      once = Boolean(optionsOrCapture.once);
-      passive = Boolean(optionsOrCapture.passive);
+    let capture, once;
+    if (
+      optionsOrCapture &&
+      (typeof optionsOrCapture === 'object' ||
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        (optionsOrCapture as any) instanceof Object)
```
justinfagnani, line 46: same here

packages/webcomponentsjs/ts_src/platform/event-listener-options.ts
```diff
@@ -0,0 +1,129 @@
...
+    const {capture, once} = parseEventOptions(options);
+    const map = getListenerMap(this, type, capture);
+    let wrapper = map.get(listener);
+    if (wrapper === undefined) {
+      wrapper = (e: Event) => {
+        map.delete(listener);
+        // Try to remove both listener and wrapper since `once` state
+        // is not used as a unique key and we don't track which was added.
+        origRemoveEventListener.call(this, type, wrapper, capture);
+        origRemoveEventListener.call(this, type, listener, capture);
```
bicknellr, line 101: Isn't `wrapper` always the right choice here, since this is happening inside `wrapper`?

packages/webcomponentsjs/ts_src/platform/event-listener-options.ts
```diff
@@ -0,0 +1,129 @@
...
+        ((listener as EventListenerObject).handleEvent ?? listener)(e);
+      };
+      map.set(listener, wrapper);
+      const nativeListener = once ? wrapper : listener;
```
bicknellr, line 105: If you added `nativeListener` to the map instead of `wrapper`, I think you could avoid calling `origRemoveEventListener` twice in remove also, since the correct listener would always be whatever comes out of the map.

bicknellr (changes requested):

packages/webcomponentsjs/ts_src/platform/event-listener-options.ts
```diff
@@ -0,0 +1,129 @@
...
+    options?: boolean | AddEventListenerOptions | undefined
+  ) {
+    if (listener == null) {
+      return;
+    }
+    const {capture, once} = parseEventOptions(options);
+    const map = getListenerMap(this, type, capture);
+    let wrapper = map.get(listener);
+    if (wrapper === undefined) {
+      wrapper = (e: Event) => {
```
bicknellr, line 96: I think this is the same
```suggestion
    if (!map.has(listener)) {
      const wrapper = (e: Event) => {
```

packages/webcomponentsjs/ts_src/platform/event-listener-options.ts
sorvell: Changed based on next suggestion.

packages/webcomponentsjs/ts_src/platform/event-listener-options.ts
sorvell: Done.

packages/webcomponentsjs/ts_src/platform/event-listener-options.ts
sorvell: Done.

packages/webcomponentsjs/ts_src/platform/event-listener-options.ts
```diff
@@ -0,0 +1,123 @@
...
+      return;
+    }
+    const {capture, once} = parseEventOptions(options);
+    const map = getListenerMap(this, type, capture);
+    if (!map.has(listener)) {
+      const cachedListener = once
+        ? (e: Event) => {
+            map.delete(listener);
+            origRemoveEventListener.call(this, type, cachedListener, capture);
+            ((listener as EventListenerObject).handleEvent ?? listener)(e);
```
bicknellr, line 99: I think this needs to be called with `call` to set up the correct `this`: `this`, when `listener` is the function being called, and `e.currentTarget`, when calling `listener.handleEvent`: if I'm reading the spec correctly, the DOM normally calls `handleEvent` as part of ['inner invoke'](https://dom.spec.whatwg.org/#concept-event-listener-inner-invoke), which calls ['call a user object's operation'](https://heycam.github.io/webidl/#call-a-user-objects-operation) by passing the event's `currentTarget` as `thisArg`. 'Call a user object's operation' also first checks if the callback object it's given is itself callable (calling that, if so) before falling back to trying to read `opName` (`"handleEvent"`).

packages/shadydom/src/patch-events.js
sorvell: Done.

packages/shadydom/src/patch-events.js
```diff
@@ -43,22 +43,23 @@ const supportsEventOptions = (() => {
 })();
 
 const parseEventOptions = (optionsOrCapture) => {
-  let capture, once, passive, shadyTarget;
-  if (optionsOrCapture && typeof optionsOrCapture === 'object') {
+  let capture, once, shadyTarget;
+  if (
+    optionsOrCapture &&
```
justinfagnani, line 48: super nit: this only value that `optionsOrCapture &&` is guarding against now is `null`. You could do an explicit check:
```suggestion
    optionsOrCapture !== null &&
```

packages/shadydom/src/patch-events.js
```diff
@@ -379,35 +403,14 @@ function shadyDispatchEvent(e) {
...
-  return (
-    node === savedNode &&
-    type === savedType &&
-    capture === savedCapture &&
-    once === savedOnce &&
-    passive === savedPassive
-  );
+function listenerSettingsEqual(savedListener, node, type, capture) {
+  let {node: savedNode, type: savedType, capture: savedCapture} = savedListener;
+  return node === savedNode && type === savedType && capture === savedCapture;
```
justinfagnani, line 408: This is probably ok, but ideally I'd want to check what the browser considers equal for options. ie, should these be loose checks?

packages/shadydom/src/patch-events.js
```diff
@@ -271,6 +272,29 @@ let EventPatches = {
   },
 };
 
+const hasDescriptors = utils.settings.hasDescriptors;
+
+const eventPhaseDescriptor =
+  hasDescriptors &&
+  Object.getOwnPropertyDescriptor(Event.prototype, 'eventPhase');
+if (eventPhaseDescriptor) {
```
justinfagnani, line 280: Why is this conditional? It's supported back to IE9. Add a comment?

packages/shadydom/src/patch-events.js
```diff
@@ -271,6 +272,29 @@ let EventPatches = {
   },
 };
 
+const hasDescriptors = utils.settings.hasDescriptors;
```
justinfagnani, line 275: Is this related to listener options?

justinfagnani (commented):
Should the eventPhase stuff be a separate PR? What was the bug here?

packages/shadydom/src/patch-events.js
sorvell: Think these have to exactly match, see https://dom.spec.whatwg.org/#add-an-event-listener.

packages/shadydom/src/patch-events.js
sorvell: Added comment. Chrome 41 (a test browser) is missing it.

packages/shadydom/src/patch-events.js
sorvell: Safari 9 has broken DOM descriptors and this catches that. Added a comment.

packages/shadydom/src/patch-events.js
justinfagnani: Are we still testing on 41?

Which supported browser does that stand in for now... older Cobalt?

sorvell: Actually no, but it's probably good to continue to test on given the range of browsers needed... internally.

bicknellr:
+1 on splitting the event phase part into a different PR, if possible.

packages/webcomponentsjs/ts_src/platform/event-listener-options.ts
```diff
@@ -0,0 +1,142 @@
...
+        ? (e: Event) => {
+            map.delete(listener);
+            origRemoveEventListener.call(this, type, cachedListener, capture);
+            ((listener as EventListenerObject).handleEvent ?? listener).call(
+              this,
+              e
+            );
+          }
+        : listener;
+      map.set(listener, cachedListener);
```
justinfagnani, line 121: I think if you add the listener as both the key and value it'll be held on to strongly. I would add another value like `undefined`

packages/webcomponentsjs/ts_src/platform/event-listener-options.ts
```diff
@@ -0,0 +1,142 @@
...
+    const map = getListenerMap(this, type, capture);
+    if (!map.has(listener)) {
+      const cachedListener = once
+        ? (e: Event) => {
+            map.delete(listener);
+            origRemoveEventListener.call(this, type, cachedListener, capture);
+            ((listener as EventListenerObject).handleEvent ?? listener).call(
+              this,
+              e
+            );
```
bicknellr, line 118: This first needs to check whether or not `listener` is callable before trying `listener.handleEvent`:

```suggestion
            if (typeof listener === 'function') {
              return listener.call(this, e);
            }

            const handleEvent = listener?.handleEvent;
            if (typeof handleEvent === 'function') {
              return handleEvent.call(e.currentTarget, e);
            }
```

packages/webcomponentsjs/ts_src/platform/event-listener-options.ts
sorvell: Done.

packages/webcomponentsjs/ts_src/platform/event-listener-options.ts
```diff
@@ -0,0 +1,142 @@
...
+      capture,
+      once,
+    };
+  };
+
+  const origAddEventListener = nativeEventTarget.prototype.addEventListener;
+  const origRemoveEventListener =
+    nativeEventTarget.prototype.removeEventListener;
+
+  const captureListenerMap = new WeakMap();
```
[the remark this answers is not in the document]
sorvell: Added types. Don't want to change the project options, but we can make an issue if it's important.

packages/tests/webcomponentsjs_/event-listener-options.html
```diff
@@ -0,0 +1,371 @@
...
+            });
+            el1.click();
+            assert.isTrue(
+              functionCalled,
+              'function called when handleEvent is defined on it'
+            );
+            assert.isFalse(
+              handleEventCalled,
+              'handleEvent not called when on listener function'
+            );
```
bicknellr, line 251: :+1:

packages/tests/webcomponentsjs_/event-listener-options.html
```diff
@@ -0,0 +1,371 @@
...
+            el1.dispatchEvent(new Event('click'));
+            assert.equal(callback.callCount, 0, 'not called if removed early');
+            el1.addEventListener('click', callback, {once: true});
+            el1.removeEventListener('click', callback, {once: true});
+            el2.addEventListener('click', callback, {once: true});
+            el1.dispatchEvent(new Event('click'));
+            assert.equal(callback.callCount, 0, 'not called if removed early');
+            el2.dispatchEvent(new Event('click'));
+            el2.dispatchEvent(new Event('click'));
+            assert.equal(callback.callCount, 1, 'not called if removed early');
```
bicknellr, line 300: The assert message feels off for this case.

packages/tests/webcomponentsjs_/event-listener-options.html
```diff
@@ -0,0 +1,371 @@
...
+              cCurrentTarget = e.currentTarget;
+            };
+            parent.addEventListener('test', pListener, {once: true});
+            child.addEventListener('test', cListener, {once: true});
+            gchild.dispatchEvent(new Event('test', {bubbles: true}));
+            assert.ok(pThisRef);
+            assert.equal(pThisRef, pCurrentTarget);
+            assert.ok(cThisRef);
+            assert.equal(cThisRef, cCurrentTarget);
+          });
```
bicknellr, line 360: Could you add one for the `handleEvent` case also?

packages/shadydom/src/patch-events.js
```diff
@@ -271,6 +272,31 @@ let EventPatches = {
   },
 };
 
+// Some browsers have missing or broken descriptors. If this is the case
+// eventPhase can't be polyfilled. For example, Chrome 41 does not
+// have prototypical descriptors on DOM objects and Safari 9 has broken ones
+// that cannot be used.
```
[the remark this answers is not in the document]
sorvell: Don't think it's worth it since this affects only these really old browsers and `eventPhase` generally isn't critically important. We'd likely need to do the manual `dispatchEvent` always on these browsers to fix it, and that doesn't seem worth it.

packages/shadydom/src/patch-events.js
```diff
@@ -271,6 +272,31 @@ let EventPatches = {
...
+const eventPhaseDescriptor =
+  utils.settings.hasDescriptors &&
+  Object.getOwnPropertyDescriptor(Event.prototype, 'eventPhase');
+if (eventPhaseDescriptor) {
+  Object.defineProperty(EventPatches, 'eventPhase', {
+    get() {
+      return this.currentTarget === this.target
```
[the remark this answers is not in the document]
sorvell: Events currently aren't patched to have `SHADY_PREFIX` properties. The patches for elements and events are treated differently.

Element prototypes are patched and in noPatch mode, they aren't patched. We access the `SHADY_PREFIX` properties because they always exist and if a noPatch wrapper calls a method, we can safely call the `SHADY_PREFIX` versions.

Event instances are patched so we will only ever get to this code on a patched instance. I can't say for sure why the Event prototype isn't patched and it's something perhaps we can revisit, but this is how it works right now.

packages/tests/shadydom/shady.html
```diff
@@ -1301,11 +1383,91 @@
...
+          );
+          assert.equal(callback.callCount, 2, 'listener not called');
+        });
+
         test('support `once` option', function () {
           ShadyDOM.wrap(el1).addEventListener('click', callback, {once: true});
-          el1.click();
+          ShadyDOM.wrapIfNeeded(el1).dispatchEvent(
+            new Event('click', {bubbles: true, composed: true})
+          );
```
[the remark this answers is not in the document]
sorvell: We need to ensure the event dispatches through ShadyDOM. In `noPatch` mode it will not. In general we use `wrap` to ensure this, but the wrapper does not currently have `click` (a separate issue).

packages/tests/webcomponentsjs_/event-listener-options.html
sorvell: Fixed assert messages.

packages/tests/shadydom/shady.html
```diff
@@ -1463,6 +1659,70 @@
...
+          ShadyDOM.wrapIfNeeded(h1_c_h2_c_h3_c).addEventListener(
+            'click',
+            listener
+          );
+          ShadyDOM.wrapIfNeeded(h1_c_h2_c_h3_c).dispatchEvent(
+            new Event('click', {bubbles: true, composed: true})
+          );
+          const t = Event.AT_TARGET;
+          const b = Event.BUBBLING_PHASE;
+          assert.deepEqual(eventPhases, [t, t, b, t, b, t]);
```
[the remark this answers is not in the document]
sorvell: It does. Added this into the tests.

packages/shadydom/src/patch-events.js
```diff
@@ -514,7 +515,7 @@ export function addEventListener(type, fnOrObj, optionsOrCapture) {
    * @this {HTMLElement}
    * @param {Event} e
    */
-  const wrapperFn = function (e) {
+  const wrapperFn = (e) => {
```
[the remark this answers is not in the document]
sorvell: Think this was just a mistake. Reverting, but the @this here **should** be EventTarget so will update that.

packages/tests/webcomponentsjs_/event-listener-options.html
sorvell: Added and this revealed an issue. I can't find it in the spec, but via empirical testing, `this` is the listener object itself when listening with an object with `handleEvent`.

bicknellr: Ok, if this is what browsers do then that sounds good.

packages/webcomponentsjs/CHANGELOG.md
```diff
@@ -8,6 +8,12 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 <!-- ## [Unreleased] -->
 
+## [Unreleased]
+
+- Polyfill `addEventListener/removeEventListener` event listener options,
+  including `{capture: boolean, once: boolean}`
+  ([#469]https://github.com/webcomponents/polyfills/pull/469)
```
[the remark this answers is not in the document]
bicknellr: just want to make sure you saw this one

packages/tests/webcomponentsjs_/event-listener-options.html
bicknellr: I misread the spec. From [call a user object's operation](https://heycam.github.io/webidl/#call-a-user-objects-operation), step 10.5:

> Set *thisArg* to *O* (overriding the provided value).

Where *O* is the listener object at that point.

packages/tests/shadydom/shady.html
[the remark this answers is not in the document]
bicknellr: Isn't `once` handled by `addEventListener` and the listener itself? I'm confused about which part a browser-dispatched event would miss.

packages/webcomponentsjs/CHANGELOG.md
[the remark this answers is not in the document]
sorvell: Looks good.

packages/tests/shadydom/shady.html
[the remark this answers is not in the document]
sorvell: If the event is not dispatched via ShadyDOM, it won't get the `composed` flag on browsers that don't support it. We depend on that flag in the listeners.

packages/webcomponentsjs/CHANGELOG.md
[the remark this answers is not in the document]
bicknellr: It needs those extra parens though.

packages/tests/shadydom/shady.html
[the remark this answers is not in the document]
bicknellr: Oh, now I get it, so this change isn't specifically related to this test but rather for the others where there are polyfilled shadow roots?

bicknellr (approved):
Other than the parens in the readme, LGTM

Issue #481, opened 2022-01-18
Scoped registry breaks formAssociated elements


calebdwilliams:
### Description

Using the scoped elements registry breaks `formAssociated` components, somehow any formAssociated getter doesn't get set on the elements.

### Example

You can find a reproduction [on this StackBlitz demo](https://stackblitz.com/edit/js-4jvsvr?file=index.js).

### Steps to reproduce

```javascript
import '@webcomponents/scoped-custom-element-registry';

class Whatever extends HTMLElement {
  static get formAssociated() {
    return true;
  }

  internals = this.attachInternals();

  connectedCallback() {
    this.internals.setFormValue('foo');
  }
}

customElements.define('what-ev', Whatever);

document.getElementById('form').addEventListener('submit', (event) => {
  event.preventDefault();
  console.log(new FormData(event.target).get('whatever'));
});

const wrapper = document.createElement('div');
wrapper.innerHTML = `<form>
  <what-ev></what-ev>
</form>`;

/**
 * The above throws the following error
 * preview-a214a06f7a20.js:2 DOMException: Failed to execute 'setFormValue' on 'ElementInternals': The target element is not a form-associated custom element.
 */
```

#### Expected behavior

Form associated behavi

### Version

0.0.3

### Browsers affected

- [x] Chrome
- [x] Firefox ([via the polyfill](https://www.npmjs.com/package/element-internals-polyfill))
- [x] Edge ([via the polyfill](https://www.npmjs.com/package/element-internals-polyfill))
- [x] Safari ([via the polyfill](https://www.npmjs.com/package/element-internals-polyfill))
- [ ] IE 11

calebdwilliams:
This is fixed in version 0.0.4. Thanks all!

PR #482, opened 2022-01-18
fix(scoped-custom-element-registry): pass formAssociated to standInEl…

calebdwilliams:
Addresses #481. The catch here is that in order to work this  changes the scoped custom element registry Closure compilation level to 'SIMPLE' which might cause extra bloat. There is probably a better way to do that, but I'm relatively unfamiliar with Closure Compiler. 

Happy to update as needed or would accept changes to this PR from maintainers. This is a pretty huge blocker for me as I'm unable to implement form-associated custom elements and use the polyfill.

### DIFF
--- a/packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js
+++ b/packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js
@@ -276,29 +276,29 @@ if (!ShadowRoot.prototype.createElement) {
       // Form-associated custom elements lifecycle methods
       formAssociatedCallback() {
         const definition = definitionForElement.get(this);
-        if (definition?.formAssociated) {
-          definition?.formAssociatedCallback?.apply(this, arguments);
+        if (definition && definition['formAssociated']) {
+          definition?.['formAssociatedCallback']?.apply(this, arguments);
         }
       }
 
       formDisabledCallback() {
         const definition = definitionForElement.get(this);
-        if (definition?.formAssociated) {
-          definition?.formDisabledCallback?.apply(this, arguments);
+        if (definition?.['formAssociated']) {
+          definition?.['formDisabledCallback']?.apply(this, arguments);
         }
       }
 
       formResetCallback() {
         const definition = definitionForElement.get(this);
-        if (definition?.formAssociated) {
-          definition?.formResetCallback?.apply(this, arguments);
+        if (definition?.['formAssociated']) {
+          definition?.['formResetCallback']?.apply(this, arguments);
         }
       }
 
       formStateRestoreCallback() {
         const definition = definitionForElement.get(this);
-        if (definition?.formAssociated) {
-          definition?.formStateRestoreCallback?.apply(this, arguments);
+        if (definition?.['formAssociated']) {
+          definition?.['formStateRestoreCallback']?.apply(this, arguments);
         }
       }
 
--- a/packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js
+++ b/packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js
@@ -74,7 +74,10 @@ if (!ShadowRoot.prototype.createElement) {
       // Register a stand-in class which will handle the registry lookup & delegation
       let standInClass = nativeGet.call(nativeRegistry, tagName);
       if (!standInClass) {
-        standInClass = createStandInElement(tagName);
+        standInClass = createStandInElement(
+          tagName,
+          elementClass.formAssociated
+        );
         nativeDefine.call(nativeRegistry, tagName, standInClass);
       }
       if (this === window.customElements) {
@@ -206,8 +209,11 @@ if (!ShadowRoot.prototype.createElement) {
 
   // Helper to create stand-in element for each tagName registered that delegates
   // out to the registry for the given element
-  const createStandInElement = (tagName) => {
+  const createStandInElement = (tagName, formAssociated) => {
     return class ScopedCustomElementBase {
+      static get formAssociated() {
+        return formAssociated;
+      }
       constructor() {
         // Create a raw HTMLElement first
         const instance = Reflect.construct(
--- a/packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js
+++ b/packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js
@@ -476,5 +476,76 @@ if (!ShadowRoot.prototype.createElement) {
         }
       };
     });
+
+    // Emulate the native RadioNodeList object
+    class RadioNodeList extends Array {
+      constructor(elements) {
+        super(...elements);
+        this._elements = elements;
+      }
+
+      get value() {
+        return this._elements.find((element) => element.checked.value);
+      }
+    }
+
+    // Emulate the native HTMLFormControlsCollection object
+    class HTMLFormControlsCollection {
+      constructor(elements) {
+        const entries = new Map();
+        elements.forEach((element, index) => {
+          const name = element.getAttribute('name');
+          this[+index] = element;
+          if (name && entries.has[name]) {
+            entries.get(name).push(element);
+          } else {
+            entries.set(name, [element]);
+          }
+        });
+        this._length = elements.length;
+        entries.forEach((value, key) => {
+          if (!value) return;
+          if (value.length === 1) {
+            this[key] = value[0];
+          } else {
+            this[key] = new RadioNodeList(value);
+          }
+        });
+      }
+
+      namedItem(key) {
+        return this[key];
+      }
+
+      get length() {
+        return this._length;
+      }
+    }
+
+    // Override the built-in HTMLFormElements.prototype.elements getter
+    const formElementsDescriptor = Object.getOwnPropertyDescriptor(
+      HTMLFormElement.prototype,
+      'elements'
+    );
+    Object.defineProperty(HTMLFormElement.prototype, 'elements', {
+      get() {
+        const nativeElements = formElementsDescriptor.get.call(this, []);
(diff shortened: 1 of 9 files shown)


calebdwilliams:
CLA signed, for what it's worth.

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js
```diff
@@ -74,7 +74,10 @@ if (!ShadowRoot.prototype.createElement) {
       // Register a stand-in class which will handle the registry lookup & delegation
       let standInClass = nativeGet.call(nativeRegistry, tagName);
       if (!standInClass) {
-        standInClass = createStandInElement(tagName);
+        standInClass = createStandInElement(
+          tagName,
+          elementClass.formAssociated
```
kevinpschaaf, line 79: So the problem with the `standInElement` is that only the settings on the first element registered will be respected by the browser.

For callbacks, the `standInElement` registers generic callbacks that can then delegate at runtime to the class for the scope it's used in. However, it's harder for static settings that can't be delegated at runtime. So for `formAssociated`, if the first class that happens to be registered for that tag name has `formAssociated: false`, any subsequent ones won't be form-associated.

There's a similar problem with `observedAttributes`, but we're able to mostly work around that by patching `setAttribute` to come close to mimicking the browser behavior for the per-class `observedAttributes` settings. It seems like we'd need to default this to `true` for all classes, but I don't know all the implications of that or if there's a way at runtime to make a `formAssociated: true` element behave as if it wasn't form associated; that would need some thought, and if it's not possible, it might just be a hard limitation of the polyfill.

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js
```diff
@@ -206,8 +209,11 @@ if (!ShadowRoot.prototype.createElement) {
 
   // Helper to create stand-in element for each tagName registered that delegates
   // out to the registry for the given element
-  const createStandInElement = (tagName) => {
+  const createStandInElement = (tagName, formAssociated) => {
     return class ScopedCustomElementBase {
+      static get formAssociated() {
```
kevinpschaaf, line 214: If the issue with `ADVANCED` compilation is that this field is being renamed (likely because closure doesn't have knowledge of the new `formAssociated field`), then doing `static get ['formAssociated']()` should fix that.

kevinpschaaf (changes requested):
So if I understand correctly, the issue in #481 is not concerned with making `formAssociated` work in scoped registries per-se, more that just loading the polyfill breaks `formAssociated` in the global registry.

The change here would certainly fix the global-registry-only usage (since the `standInElement` will be the only element registered, so the `formAssociated` setting will always be correct). However, it will be incorrect for scoped registries.

This needs a bit more research as to whether we can make a scheme to use the browser's built-in `formAssociated` setting for all registered classes, and defeat its "form-associatedness" at runtime somehow for classes that had `formAssociated: false`.

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js
calebdwilliams: re: callbacks, does that mean we'd need to add something to watch for the new callbacks like `formAssociatedCallback`, `formResetCallback`, etc?

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js
calebdwilliams: Doing that results in `[JSC_ILLEGAL_PROPERTY_ACCESS] Cannot do '[]' access on a struct`

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js
```diff
@@ -209,10 +212,10 @@ if (!ShadowRoot.prototype.createElement) {
 
   // Helper to create stand-in element for each tagName registered that delegates
   // out to the registry for the given element
-  const createStandInElement = (tagName, formAssociated) => {
+  const createStandInElement = (tagName) => {
     return class ScopedCustomElementBase {
-      static get formAssociated() {
-        return formAssociated;
+      static get ['formAssociated']() {
+        return true;
```
kevinpschaaf, line 218: So one other thing we should consider whether we can fix: Another side-effect of setting `true` here is that an element will show up in `form.elements`. I wonder if `Form.prototype.elements` is patchable to filter out CE's with `!formAssociated`....

calebdwilliams: Haha, try as I may I cannot patch `Form.prototype.elements`, that's an [open issue on the polyfill right now](https://github.com/calebdwilliams/element-internals-polyfill/issues/43).

calebdwilliams: TIL, an element is added to elements if `formAssociated` is true even if internals isn't set …

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js
```diff
@@ -269,6 +272,27 @@ if (!ShadowRoot.prototype.createElement) {
         definition?.adoptedCallback?.apply(this, arguments);
       }
 
+      // Form-associated custom elements lifecycle methods
+      formAssociatedCallback() {
+        const definition = definitionForElement.get(this);
+        definition?.formAssociatedCallback?.apply(this, arguments);
```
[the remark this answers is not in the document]
calebdwilliams: Good call.

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js
kevinpschaaf: I was thinking something like this: https://lit.dev/playground/?mods=gists#gist=be7e360bf80fb6431c1f5dc532063f4f

(the commented out code is roughly what I was thinking... the uncommented `return []` is just to prove that the `Form.prototype.elements` is indeed patchable...)

calebdwilliams: @kevinpschaaf yeah, I've got a POC working, but one test is failing for some reason and there'll need to be several more tests written, but it's nearly there.

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js
```diff
@@ -403,4 +441,111 @@ if (!ShadowRoot.prototype.createElement) {
...
+
+    // Emulate the native RadioNodeList object
+    class RadioNodeList extends Array {
+      constructor(elements) {
+        super(...elements);
+        this._elements = elements;
+      }
+
+      get value() {
+        return this._elements.find((element) => element.checked.value);
```
kevinpschaaf, line 490: I think you want to dereference value off of the found checked one, right? And looks like it defaults to `''` when there's nothing selected.
```suggestion
        return this._elements.find((element) => element.checked)?.value ?? '';
```

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js
```diff
@@ -403,4 +442,110 @@ if (!ShadowRoot.prototype.createElement) {
...
+
+        const include = [];
+
+        for (const element of nativeElements) {
+          const definition = definitionForElement.get(element);
+
+          // Only purposefully formAssociated elements or built-ins will feature in elements
+          if (
+            (definition && definition['formAssociated'] === true) ||
+            element.localName.indexOf('0') === -1
```
kevinpschaaf, line 541: But, I guess as long as it doesn't have a `definition`, it's ok to let an element filter through, probably don't need the tag name check at all. So maybe just
```suggestion
            !definition || definition['formAssociated'] === true
```

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js
kevinpschaaf: And would you mind adding a test for the radio group case? Looks like this would have thrown.

kevinpschaaf (changes requested):
Awesome work on this! I think it's really close, just a couple of minor fixes I think. Nice job on paying attention to the `HTMLFormControlsCollection` detail there.

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js
```diff
@@ -403,4 +441,111 @@ if (!ShadowRoot.prototype.createElement) {
...
+        const nativeElements = formElementsDescriptor.get.call(this, []);
+        const include = [];
+
+        for (const element of nativeElements) {
+          const definition = definitionForElement.get(element);
+
+          // Only purposefully formAssociated elements or built-ins will feature in elements
+          if (
+            definition?.formAssociated === true ||
+            element.localName.indexOf('0') === -1
```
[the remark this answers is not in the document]
kevinpschaaf: ^^ suggestion probably obsoleted by suggestion below

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js
calebdwilliams: Done!

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js
calebdwilliams: Added.

calebdwilliams:
I made the requested changes, hopefully this commit will do it. Do you want me to rebase or is it fine having nine commits on this PR?

kevinpschaaf (approved):
LGTM

PR #484, opened 2022-01-20
Shady ondemand fixes

sorvell:
* adds `ShadyDOM.validateNodePatch` which can reset a patch on a node if needed.
* can specify `ShadyDOM.onDemandPatches`, a list of properties to always patch when in on-demand mode.
* always patches styling related properties (`setAttribute`, `className`) on elements
* allows a limited form of overriding native methods in custom elements by copying implemented methods to the "native" version

### DIFF
--- a/packages/shadydom/src/patch-prototypes.js
+++ b/packages/shadydom/src/patch-prototypes.js
@@ -108,19 +108,22 @@ const getPatchPrototype = (name) => window[name] && window[name].prototype;
 // accessors.
 const disallowedNativePatches = utils.settings.hasDescriptors
   ? null
-  : ['innerHTML', 'textContent'];
+  : {innerHTML: true, textContent: true};
 
 /**
  * Patch a group of accessors on an object.
  * @param {!Object} proto
  * @param {!Array<Object>} list
  * @param {string=} prefix
- * @param {Array=} disallowed
+ * @param {Object=} disallowed
+ * @param {Object=} allowed
  */
-function applyPatchList(proto, list, prefix, disallowed) {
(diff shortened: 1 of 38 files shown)


packages/shadydom/src/patch-prototypes.js
```diff
@@ -196,7 +199,12 @@ export const patchElementProto = (proto, isVanillaElement = false) => {
     copyAllImplementedNativeProperties(proto);
   }
   proto[PROTO_IS_PATCHED] = true;
-  applyAllHTMLElementProtoPatches(proto);
+  allHTMLElementOnDemandPatches.forEach((patches) => {
```
kevinpschaaf, line 202: Question (can't recall): We may need to on-demand non-element (TextNode, CommentNode, etc.); do we just patch in all of the HTMLElement patches onto those nodes and call it good enough? Or is the node type being patched discriminated somewhere?

packages/shadydom/src/utils.js
```diff
@@ -252,13 +252,21 @@ export const patchExistingProperties = (proto, descriptors) => {
...
  */
-export const copyImplementedNativeProperties = (proto, descriptors) => {
+export const copyImplementedNativeProperties = (
+  proto,
+  descriptors,
+  disallowed
+) => {
   for (let name in descriptors) {
     const d = descriptors[name];
     // Only check methods.
```
kevinpschaaf, line 262: Why not also give this treatment for accessors? Might be nice to not need to come in and add it later if we decide we need a className override or something...? 🤷

kevinpschaaf (approved):
LGTM, a couple things to consider

PR #486, opened 2022-01-25
fix(scoped-custom-element-registry): remove square brackets from method names

bgotink:
#482 contained a couple of typos that break certain form-associated callbacks. Instead of putting these three names inside square brackets to prevent closure from mangling them, the square brackets were placed inside the method names.


aomarks (approved):
That seems bad! I guess we don't have test coverage for this?

kevinpschaaf (approved):

PR #501, opened 2022-03-31
Make `webcomponents-loader.js` compatible with the Trusted Types API

bicknellr:
Fixes https://github.com/Polymer/internal/issues/1272


bicknellr:
Local WCT tests passed as expected, since no trusted types policy being enforced yet:
https://github.com/webcomponents/polyfills/runs/5763342598?check_suite_focus=true#step:7:24

bicknellr:
After adding the new policy, the tests pass again:
https://github.com/webcomponents/polyfills/runs/5763782787?check_suite_focus=true#step:7:24

packages/webcomponentsjs/webcomponents-loader.js
```diff
@@ -157,16 +157,54 @@
...
+      if (
+        window.trustedTypes &&
+        trustedTypesAreEnforced &&
+        !window.trustedTypes.isScriptURL(window.WebComponents.root)
+      ) {
+        throw new Error('`WebComponents.root` must be a `TrustedScriptURL`.');
+      }
+      url = trustedTypesPolicyWrapper.createScriptURL(
+        window.WebComponents.root + polyfillFile
+      );
```
bicknellr, line 201: @koto, I'm curious what you think about this w.r.t. https://github.com/w3c/webappsec-trusted-types/issues/36. The issue here is that `WebComponents.root` is arbitrary input used to build a string that I'm stamping here as a `TrustedScriptURL`, so I'd like to enforce that it is also a `TrustedScriptURL`. However, if I don't check for and enforce only at the same time as the browser, then this error would be thrown even if the browser is only reporting trusted type violations. Is there a better way to propagate enforcement like this?

koto: We think it's an antipattern to change the behavior only if `require-trusted-types-for` (or any CSP directive, for that matter) is enforced. The issue with that is that report-only can't be safely used to iron out all the violations, and authors risk breaking production apps the moment enforcing CSP replaces the reporting one. 

The safe (but breaking) route out of this is, I think, to introduce `WebComponents.rootAsTrustedScriptURL` (naming TBD) and use only that when `window.trustedTypes` is available. 

The less-safe (but b/w compatible) is to continue using `.root`, and just documenting that the value should not be user controlled. To my understanding the loader is only used for polyfills anyway, and these would only be loaded for older browsers, which could only get TT via a polyfill, so the practical risk is minimal here.

I would personally just go for the second option, given the constraints.

packages/webcomponentsjs/README.md
```diff
@@ -157,6 +161,15 @@ Here's a more complicated example:
...
 ```
 
+#### Trusted Types
+
+If you're using the loader on a page that enforces the `trusted-types` Content
+Security Policy, you'll need to allow the `webcomponents-loader` policy name so
+that the loader can dynamically create and insert a `<script>` for the polyfill
+bundle it selects based on feature detection. You should also treat
+`WebComponents.root` as if it were required to be set to a `TrustedScriptURL`
+when trusted types are being enforced.
```
rictic, line 171:
```suggestion
bundle it selects based on feature detection. If you set `WebComponents.root`
(which is rare), it should be set to a TrustedScriptURL for Trusted Types
compatibility.
```

packages/webcomponentsjs/webcomponents-loader.js
```diff
@@ -157,16 +157,42 @@
...
+      return policy || policyOptions;
+    })();
+
     var url;
     var polyfillFile = 'bundles/webcomponents-' + polyfills.join('-') + '.js';
 
     // Load it from the right place.
     if (window.WebComponents.root) {
-      url = window.WebComponents.root + polyfillFile;
+      url = policy.createScriptURL(window.WebComponents.root + polyfillFile);
```
rictic, line 191: This doesn't check that Webcomponents.root is a TrustedScriptUrl, but the README suggested that it would

If we want to check that, I think the code would be roughly:

```javascript
url = window.WebComponents.root + polyfillFile;
if (window.trustedTypes && 
    window.trustedTypes.isScriptURL(window.WebComponents.root)) {
  url = policy.createScriptURL(url);
}
```

wct-trusted-types-plugin/README.md
```diff
@@ -0,0 +1,29 @@
+# WCT Trusted Types Plugin
```
rictic, line 1: Note that you can also set the CSP headers with <meta> tags, which might be easier.

See: https://lit.dev/playground/#project=W3sibmFtZSI6ImluZGV4Lmh0bWwiLCJjb250ZW50IjoiPCFET0NUWVBFIGh0bWw-XG48aGVhZD5cbiAgPG1ldGEgaHR0cC1lcXVpdj1cImNvbnRlbnQtc2VjdXJpdHktcG9saWN5XCIgY29udGVudD1cInJlcXVpcmUtdHJ1c3RlZC10eXBlcy1mb3IgJ3NjcmlwdCdcIiAvPlxuPC9oZWFkPlxuPGJvZHk-XG48c2NyaXB0PlxuICBjb25zdCBlbmZvcmNlZCA9ICgoKSA9PiB7XG4gICAgdHJ5IHtcbiAgICAgIGRvY3VtZW50LmNyZWF0ZUVsZW1lbnQoJ3AnKS5pbm5lckhUTUwgPSAnJztcbiAgICAgIHJldHVybiBmYWxzZTtcbiAgICB9IGNhdGNoIHtcbiAgICAgIHJldHVybiB0cnVlO1xuICAgIH1cbiAgfSkoKTtcbiAgZG9jdW1lbnQuYm9keS5hcHBlbmRDaGlsZChuZXcgVGV4dChgVHJ1c3RlZCB0eXBlcyAke2VuZm9yY2VkID8gJ2FyZScgOiAnYXJlIG5vdCd9IGVuZm9yY2VkIG9uIHRoaXMgcGFnZWApKTtcbjwvc2NyaXB0PlxuPC9ib2R5PlxuIn0seyJuYW1lIjoicGFja2FnZS5qc29uIiwiY29udGVudCI6IntcbiAgXCJkZXBlbmRlbmNpZXNcIjoge1xuICAgIFwibGl0XCI6IFwiXjIuMC4wXCIsXG4gICAgXCJAbGl0L3JlYWN0aXZlLWVsZW1lbnRcIjogXCJeMS4wLjBcIixcbiAgICBcImxpdC1lbGVtZW50XCI6IFwiXjMuMC4wXCIsXG4gICAgXCJsaXQtaHRtbFwiOiBcIl4yLjAuMFwiXG4gIH1cbn0iLCJoaWRkZW4iOnRydWV9XQ

rictic (changes requested):

packages/webcomponentsjs/webcomponents-loader.js
bicknellr: It doesn't because it would be a breaking change to always check if it were one if `window.trustedTypes` exists. However, @koto [says it's considered an antipattern to detect whether or not trusted types are being enforced](https://github.com/webcomponents/polyfills/pull/501#discussion_r846735919), which would be necessary to propagate both the requirement that `WebComponents.root` be a `TrustedScriptURL` without throwing an error if trusted types are being reported, but not enforced.

packages/webcomponentsjs/README.md
bicknellr: I'll work on the language more here. I don't think we can actually enforce that this is a `TrustedScriptURL` without making a breaking change, but I think I could word this so that the idea that it should only be set to values that would be considered as safe as a `TrustedScriptURL` was more clear. I.e. we can't actually enforce the safety ourselves, so users should know that it needs to be treated this way.

wct-trusted-types-plugin/README.md
bicknellr: Oh, I totally forgot about that! I'll remove the plugin.

packages/webcomponentsjs/README.md
bicknellr: I've updated this slightly, PTAL

wct-trusted-types-plugin/README.md
bicknellr: Removed.

packages/webcomponentsjs/webcomponents-loader.js
rictic: If we instead only make url a TrustedScriptURL if WebComponents.root is a TrustedScriptUrl then existing users get the current behavior, and that's only a problem if they've opted into trusted types.

But if they've opted into trusted types, then they _want_ to be told about potential security issues, and so if they're using window.WebComponents.root then they _want to know_ that it's a potential security issue, and it should only be set to a trusted script url.

bicknellr: Ok, I now buy that this both isn't a breaking change and propagates safety since they can't use the loader with trusted types enabled currently.

bicknellr: Updated.

packages/webcomponentsjs/README.md
rictic: WDYT of my recommended wording here?

bicknellr: Yeah, this is better - updated.

rictic (approved):

bicknellr:
This is passing locally and in internal test runs so I'm going to merge despite the test infra problems.

PR #502, opened 2022-04-07
Fix focus/blur events to be usable without wrapped addEventListener

kevinpschaaf:
When `focus` and `blur` are dispatched, a global capturing event listener sees the event and manually dispatches it through the tree to listeners added with wrapped `addEventListener` calls. The manual dispatching patches `currentTarget` to return retargeted values (along with `eventPhase`, since we are simulating the bubbling phase via a capturing listener).  When all `addEventListener` calls are wrapped this works fine. 

However, because the capturing listener does not stop propagation, the event will also natively fire to any native listeners added with an unwrapped `addEventListener` call (note that Polymer [does not wrap](https://github.com/Polymer/polymer/blob/master/lib/mixins/template-stamp.js#L555) `addEventListener` for listeners added via templates when in `noPatch` mode, as a performance optimization/correctness tradeoff). Because the manual dispatching code leaves the  patched `currentTarget` accessor installed, it can cause the subsequent native listeners to see the same event, but with patched accessors that return invalid values (like `null`), rather than just the unpatched values.

This change restores the original accessors for those fields when the manual dispatching is complete, so that native listeners see native fields on the event.

Fixes https://github.com/webcomponents/polyfills/issues/503.

### DIFF
--- a/packages/shadydom/CHANGELOG.md
+++ b/packages/shadydom/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+- Fix focus/blur events to be usable without wrapped `addEventListener` in
+  `noPatch` mode. ([#502](https://github.com/webcomponents/polyfills/pull/502))
 - Add type annotations for JSCompiler to `eventPhase` property descriptor.
   ([#473](https://github.com/webcomponents/polyfills/pull/473))
 - Allow event listener options to be specified using a function in addition to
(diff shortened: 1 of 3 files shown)


aomarks (approved):

bicknellr:
This seems good too me. I think since we're having trouble with the test infra right now, an internal run on a CL that imports this branch would be a good substitute.

bicknellr (approved):
@kevinpschaaf imported this as http://cl/440345428 and the TGP results were good (2022-04-08 11:20). Also, I just ran the tests locally and can confirm that Chrome is fine despite the flakiness we've been having with the automated tests lately. LGTM

kevinpschaaf:
Manually running tests on IE also pass: 
image

PR #509, opened 2022-04-25
Migrate the custom element polyfill tests to `@web/test-runner`.

bicknellr:
This PR migrates the custom element polyfill tests from WCT to `@web/test-runner`.

- The existing scoped custom elements polyfill tests were already using WTR, but this PR does not merge them into the `tests` package.
- WTR and related dependencies have been moved from the root to the individual packages that actually depend on them (`tests` and `scoped-custom-element-registry`).
- The `tests` package now includes a plugin for `@web/dev-server` that matches enough of the compilation behavior of `polyserve` to provide the tests with the same environment as before.
- The GitHub Actions jobs for WCT and WTR tests have been merged.

---

- [x] https://github.com/modernweb-dev/web/issues/1986


bicknellr:
This still depends on getting the issue linked in the description fixed and I'd like to factor the WTR configs a bit differently, but I think this PR is close enough to start getting some feedback. PTAL, thanks!

packages/tests/package.json
```diff
@@ -2,21 +2,34 @@
   "name": "tests",
   "private": true,
   "scripts": {
-    "build": "gulp --gulpfile ./shadycss/gulpfile.js",
-    "build:watch": "chokidar --initial --follow-symlinks 'shadycss/module/**/*.js' --ignore 'shadycss/module/generated/**/*' 'node_modules/@webcomponents/shadycss/**/*.js' -c 'npm run build'"
+    "build": "gulp --gulpfile ./shadycss/gulpfile.js && rollup --config rollup.config.js",
```
rictic, line 5: Feel free to punt to a follow up PR, but we should consider using wireit here.

Maybe esbuild as well, unless rollup's output is part of the system under test

packages/tests/shadydom/event-path.html
```diff
@@ -1194,12 +1194,13 @@
...
           });
 
-          test('composedPath of events from non-Node subclasses works', function () {
+          test('composedPath of events from non-Node subclasses works', function (done) {
             const xhr = new XMLHttpRequest();
             xhr.open('GET', window.location.href);
             xhr.addEventListener('load', (e) => {
               assert.doesNotThrow(() => {
                 e.composedPath();
+                done();
```
rictic, line 1203: eep, good catch! this was basically a noop test before this change

packages/tests/web-test-runner.config.js
```diff
@@ -0,0 +1,162 @@
...
+    // These browsers only support the older JWP protocol.
+    //
+    // Sauce Labs determines which protocol to use to talk to the launched
+    // browser based on whether or not an option called `sauce:options` exists
+    // in the given configuration. If this key is passed, the newer W3C standard
+    // WebDriver protocol is used; otherwise, JWP is used.[^1]
+    // `@web/test-runner-saucelabs` adds the `sauce:options` key if it sees the
+    // `browserVersion` key, which is only supported by Sauce Labs with the
+    // standard WebDriver protocol.[^2] So, we use the older key names here so
+    // that JWP will be used with these browsers.
```
rictic, line 63: really great documentation throughout this PR, this is some very subtle stuff

packages/tests/webcomponentsjs_/wct-config.js
```diff
@@ -1,4 +1,7 @@
 window.WCT = {
+  // Force the 'root' path to be automatically detected relative to the
+  // 'wct-browser-legacy/browser.js' script.
+  root: null,
```
rictic, line 4: is this just making explicit what was implicitly the case before?

rictic (changes requested):
What do you make of these errors? We got the same error in both sauce runs:

https://github.com/webcomponents/polyfills/runs/7102848818?check_suite_focus=true#step:7:2552

https://github.com/webcomponents/polyfills/runs/7102848818?check_suite_focus=true#step:7:2552

packages/tests/package.json
bicknellr: Yeah, I think moving from lerna to wireit + npm workspaces would be a good idea for a later PR. And no, the choice of Rollup isn't critical for the tests package. It's also used to build the scoped custom element registry polyfill and I haven't looked at that one very closely but I doubt the choice matters there either. However, some of the polyfills use Closure for type checking and as a smoke test for compatibility with internal build infra, so I'd have to spend more time digging into that to know if / how much of what Closure does here could be replaced by esbuild.

bicknellr (commented):
BTW those links are to the same run :) That particular error is due to <https://github.com/modernweb-dev/web/issues/1986>. I've been locally testing this PR on Safari 9 and 10 on Sauce by modifying files my `tests` package's `node_modules` to match [the PR I pushed for that issue](https://github.com/modernweb-dev/web/pull/1993). Getting that fixed is definitely a blocker.

Also, that test in the `event-path.html` file where I added `done()` and one other in that same file fail in at least Chrome 41 when Shady DOM's `noPatch` mode is enabled. Though, I don't think these should necessarily be considered a blocker for this PR since they're legit failures AFAICT.

bicknellr:
Now that <https://github.com/modernweb-dev/web/pull/1993> has landed and those components were released in <https://github.com/modernweb-dev/web/pull/1997>, the proper way to update this would be to wait for a release of `@web/test-runner-saucelabs` that depends on at least `@web/test-runner-webdriver@0.5.1` and have the `tests` package depend on at least that version. In the mean time, I've added `@web/test-runner-webdriver@^0.5.1` as a dependency of the `tests` package, which is compatible with `@web/test-runner-saucelabs`' dependency/constraint of `@web/test-runner-webdriver@^0.5.0`, so `@web/test-runner-webdriver@0.5.1` will be installed first under `tests` and then `@web/test-runner-saucelabs` will have its dependency of `@web/test-runner-webdriver@^0.5.0` deduped against it, causing it to use `0.5.1`.

packages/tests/custom-elements/html/shim.test.html
```diff
@@ -10,214 +10,109 @@
 -->
 <title>Custom Elements Tests</title>
 <meta charset="utf-8" />
+<script src="../../node_modules/@webcomponents/webcomponentsjs/bundles/webcomponents-pf_js.js"></script>
```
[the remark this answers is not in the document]
bicknellr: Seems like this was necessary. The custom elements polyfill depends on `Promise` support (so every other test file already includes it) but this one originally didn't because it's for the native shim. The test environment now also depends on promises.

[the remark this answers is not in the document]
bicknellr: I think it would be more proper for the test environment to include a promise polyfill itself and for this one import to be removed, but I'm getting a different set of errors in IE 11 with core-js' promise polyfill added. I'll keep trying for a little bit - maybe I've chosen an entry point that assumes other features that aren't available in IE 11.

bicknellr:
Opened <https://github.com/modernweb-dev/web/pull/1999> to bump `@web/test-runner-saucelabs`' dependency range for `@web/test-runner-webdriver` to `^0.5.1`.

rictic:
Quoting the changelog presubmit:

> Please add a CHANGELOG entry, or add the "Skip Changelog" label if not required.

rictic (changes requested):
Looks like tests are still failing on Sauce

bicknellr:
I added the label because this PR has no user-facing changes. Also, those failures on Saucelabs are real failures from before this PR. In `packages/tests/shadydom/event-path.html`, the "composedPath of events from non-Node subclasses works" now fails because of the change I made here that causes the test to recognize an error thrown before the test ends. I think the other, "focus currentTarget is valid when unwrapped in noPatch", was already failing before but wasn't caught because we couldn't get the tests to run reliably and decided to rely on the internal tests instead.

rictic (approved):
Great work!

bicknellr:
Thanks for the taking the time to review!

PR #514, opened 2022-07-18
Fix document.cloneNode

Aierie:
Node patch affects `document` too, which means `this.ownerDocument.__CE_registry` will cause an error because `document.ownerDocument` is `null`.


packages/custom-elements/ts_src/Patch/Node.ts
```diff
@@ -92,9 +92,14 @@ export default function (internals: CustomElementInternals) {
 
   Node.prototype.cloneNode = function (this: Node, deep) {
     const clone = Native.Node_cloneNode.call(this, !!deep);
+    // document.ownerDocument is null
+    const registry = this.ownerDocument
+      ? this.ownerDocument!.__CE_registry
+      : (this as Document).__CE_registry;
```
bicknellr, line 98: I think this could be simplified to `(this.ownerDocument ?? this).__CE_registry` and kept inline.

Aierie: Done, thanks

omnoms:
Seems like the pull request failed a test, can this be picked up anew ? We are also facing this issue.

justinfagnani:
Is there a test we can add here?

PR #517, opened 2022-07-29
[ShadyDOM] Add `ShadyDOM.querySelectorImplementation` setting.

### LINKED ISSUE #480: ShadyDOM breaks querySelector polyfills
### Description

`querySelector` and friends require polyfills for features like `:scope` or `:has()` in older browsers.

These polyfills break when also including `@webcomponents/shadydom/shadydom.min.js`.
Not sure why as I am not familiar with the code base here.
Could be because an unpolyfilled `querySelector` is memoized or because it is overwritten without wrapping and extending. the original 

### Example

include :

- https://github.com/jonathantneal/element-qsa-scope
- @webcomponents/shadydom/shadydom.min.js

run `someElement.querySelector(':scope')`

This will not return the expected result.

#### Expected behavior

Expected ShadyDOM to leave `querySelector` as is or correctly wrap it.

#### Actual behavior

ShadyDOM breaks `querySelector` polyfills.

### Version

`"@webcomponents/shadydom": "^1.9.0"`

### Browsers affected

- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari
- [x] IE 11

bicknellr:
This PR splits the selectors passed to `querySelector` and `querySelectorAll`, then traverses the logical tree manually to match them. This fixes `:scope` by allowing Shady DOM to decide if the candidate elements being tested against simple selectors containing `:scope` are the context element and should therefore match.

- [x] ~~Only walk ancestors and inclusive descendants of the context element.~~
- [x] Iterate simple selectors in reverse to avoid walking large trees for descendant combinators.
- [x] Fix the result order by matching all complex selectors simultaneously.
- [x] Add a setting that allows switching between the current implementation, one which calls directly into the native functions, and this implementation, along with documentation explaining why this is configurable and the tradeoffs of each.
- [x] Add tests.
- [x] Add tests specifically for behavior around selector lists with multiple, top-level complex selectors. (`querySelectorAll` result ordering, parsing)

Fixes #480.


packages/shadydom/src/patches/ParentNode.js
```diff
@@ -128,24 +131,248 @@ export const ParentNodePatches = utils.getOwnPropertyDescriptors({
...
+        'selectors': compoundSelectors,
+        'joiners': combinators,
+      } = splitSelectorBlocks(complexSelector);
+
+      return {
+        compoundSelectors,
+        combinators,
+      };
+    })
+    .filter(({compoundSelectors}) => compoundSelectors.length > 0);
```
sorvell, line 171: For what case is this filter needed?

packages/shadydom/src/patches/ParentNode.js
```diff
@@ -128,24 +131,248 @@ export const ParentNodePatches = utils.getOwnPropertyDescriptors({
...
+   * @type {!Array<!SelectorMatchingCursor>}
+   */
+  let cursors = utils.flat(
+    query(contextElement, (_element) => true).map((element) => {
+      return utils.flat(
+        complexSelectors.map((complexSelectorParts) => {
+          const {compoundSelectors} = complexSelectorParts;
+          const index = compoundSelectors.length - 1;
+          if (matchesCompoundSelector(element, compoundSelectors[index])) {
+            return [
```
sorvell, line 231: Why does this need to be an array?

packages/shadydom/src/patches/ParentNode.js
```diff
@@ -128,24 +131,248 @@ export const ParentNodePatches = utils.getOwnPropertyDescriptors({
...
+
+          // For `a b`, where existing cursors have `position`s matching `b`,
+          // the candidates to test against `a` are all ancestors each cursor's
+          // `position`.
+          for (
+            let ancestor = position[utils.SHADY_PREFIX + 'parentNode'];
+            ancestor && ancestor instanceof Element;
+            ancestor = ancestor[utils.SHADY_PREFIX + 'parentNode']
+          ) {
+            if (matchesCompoundSelector(ancestor, compoundSelector)) {
```
sorvell, line 280: It wasn't obvious to me why results would ever be an array so maybe this should be commented, but the reason is that there may be > 1 candidate in the ancestor tree...

e.g.: given `c c c d` and `<c><c><c><d>`: the first cursor will be for `d` and next we need a cursor for *each* `c`, but eventually the one for the top most `c` will survive. Is that right?

sorvell (changes requested):
Looking really good. We need:
* 3 modes: (a) existing impl, (b) native QSA + filtering, (c) this impl.
* tests for each mode highlighting what works/fails in each.

packages/shadydom/src/patches/ParentNode.js
```diff
@@ -128,24 +131,178 @@ export const ParentNodePatches = utils.getOwnPropertyDescriptors({
...
+  });
+};
+
+/**
+ * @param {!Element} contextElement
+ * @param {string} selectorList
+ * @return {!Array<!Element>}
+ */
+const logicalQuerySelectorList = (contextElement, selectorList) => {
+  return deduplicateAndFilterToDescendants(
```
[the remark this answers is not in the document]
bicknellr: I think this comment was on an older commit, but yeah this is now how the final deduplication is handled.

packages/shadydom/src/patches/ParentNode.js
bicknellr: This probably isn't necessary. I just haven't tested how the library responds to all whitespace / the empty string yet.

packages/shadydom/src/patches/ParentNode.js
bicknellr: Yeah, each ancestor matching `c` needs a cursor because any of them could potentially be part of the match when it's only gotten around to testing the rightmost `c` in the selector. The cursors that point at the shallower `<c>`s will eventually be unable to find enough candidate `<c>` ancestors to match the rest of the selector and will be replaced with zero cursors.

That particular case will look something like this:
```
<c>
  <c>
    <c>
      <d> <!-- cursor[0] `c c c (d)`-->
```
```
<c> <!-- candidate for: cursor[0] `c c c( )d` -->
  <c> <!-- candidate for: cursor[0] `c c c( )d` -->
    <c> <!-- candidate for: cursor[0] `c c c( )d` -->
      <d>
```
```
<c> <!-- cursor[2] `c c (c) d` -->
  <c> <!-- cursor[1] `c c (c) d` -->
    <c> <!-- cursor[0] `c c (c) d` -->
      <d>
```
During this step, `cursor[2]` doesn't find any candidates:
```
<c> <!-- candidate for: cursor[0] `c c( )c d`, cursor[1] `c c( )c d` -->
  <c> <!-- candidate for: cursor[0] `c c( )c d` -->
    <c>
      <d>
```
```
<c> <!-- cursor[1] `c (c) c d`, cursor[2] `c (c) c d` -->
  <c> <!-- cursor[0] `c (c) c d` -->
    <c>
      <d>
```
During this step, `cursor[1]` doesn't find any candidates:
```
<c> <!-- candidate for: cursor[0] `c( )c c d` -->
  <c>
    <c>
      <d>
```
```
<c> <!-- cursor[0]: `(c) c c d` -->
  <c>
    <c>
      <d>
```

bicknellr: > The cursors that point at the shallower `<c>`s will eventually be unable to find enough candidate `<c>` ancestors to match the rest of the selector and will be replaced with zero cursors.

Or, maybe I should have said "Cursors with 'algorithmic ancestors' that pointed at the shallower `<c>`s will be unable..." since only cursors that have already succeeded in making a complete match match actually 'survive' between iterations - all others are replaced with zero or more new cursors. The overlapping tree / lineage terminology that's shared here between the DOM and the timeline of cursors makes describing this a bit difficult. :/

packages/shadydom/src/patches/ParentNode.js
bicknellr: This probably doesn't need to be.

packages/shadydom/src/patches/css-selector-splitter.js
```diff
@@ -0,0 +1,140 @@
+/**
```
romainmenke, line 1: This package doesn't handle escape sequences :

```js
console.log('a\\,b,c', splitSelector('a\\,b,c'))
// a\,b,c { selectors: [ 'a\\', 'b', 'c' ], joiners: [ ',', ',' ] }
```

romainmenke: This should work better : https://github.com/mrhenry/core-web/blob/main/packages/core-web-generator/polyfills/%7Eelement-qsa-has.js#L270-L378

packages/shadydom/src/patches/ParentNode.js
bicknellr: Removed.

packages/shadydom/src/patches/ParentNode.js
bicknellr: Removed this one and a few others.

packages/shadydom/src/shadydom.js
```diff
@@ -118,6 +118,81 @@ if (utils.settings.inUse) {
     'nativeMethods': nativeMethods,
     'nativeTree': nativeTree,
     'patchElementProto': patchElementProto,
+    // Use this setting to choose the implementation of `querySelector` and
+    // `querySelectorAll`. The logical tree exposed by Shady DOM via the DOM
```
sorvell, line 122: Maybe add: "The options provide different trade offs between accuracy and performance, as explained below."

packages/shadydom/src/shadydom.js
```diff
@@ -118,6 +118,81 @@ if (utils.settings.inUse) {
...
+    //   wrapper call.
+    //
+    //   However, the parser used for this option is not strictly correct.
+    //   Notable caveats include:
+    //
+    //     - The parser assumes the selector list it's given is syntactically
+    //     correct. When given a selector that would not parse under a full
+    //     implementation, this implementation may fail silently.
+    //
+    //     - Only top-level combinators and compound selectors are taken into
```
sorvell, line 176: Maybe change this to "contextual combinators are not supported"?

packages/tests/shadydom/shady-dynamic.html
```diff
@@ -2906,6 +3105,352 @@
           document.body.removeChild(d);
         });
       });
+
+      suite('querySelectorImplementation', function () {
```
sorvell, line 3109: * Make sure to also add asserts for `querySelector` in addition to `querySelectorAll`.

packages/tests/shadydom/runner.html
```diff
@@ -26,6 +26,12 @@
     'shady-dynamic.html?forceCustomElements=true',
     'shady-dynamic.html?noPatch=true&forceCustomElements=true',
     'shady-dynamic.html?noPatch=on-demand&forceCustomElements=true',
+    'shady-dynamic.html?querySelectorImplementation=selectorEngine',
```
sorvell, line 29: Are there tests for `querySelectorImplementation=native`?

sorvell (changes requested):

packages/shadydom/src/shadydom.js
bicknellr: I added that sentence and reworded this paragraph a bit.

packages/shadydom/src/shadydom.js
bicknellr: Yeah, this was a bit hard to describe since there's not really a standard term for this. 'Contextual combinators' is a bit confusing to me also, so I reworded it using 'nested selectors'. Do you think it's clearer after this update?

packages/shadydom/src/utils.js
```diff
@@ -291,3 +299,54 @@ export const convertNodesIntoANode = (...args) => {
...
+  }
+
+  return array;
+};
+
+/**
+ * Deduplicates items in an array.
+ *
+ * This function could normally be implemented as merely `Array.from(new
+ * Set(...))`. However, in IE 11, `Set` does not support being constructed with
```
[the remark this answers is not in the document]
bicknellr: There's also the `Array.from` issue. Do you think this is worth the complexity of checking if we're in the good case?

packages/tests/shadydom/runner.html
bicknellr: No, I've just added those though. I had to update some of them since they were doing things that aren't supported with that option.

packages/tests/shadydom/shady-dynamic.html
bicknellr: Ok, I added them for the tests for the tree-walking parts, but I left them out of the `parseSelectorList` tests since those are only to check if they parsed properly.

packages/shadydom/src/patches/logicalQuerySelectorAll.md
```diff
@@ -0,0 +1,204 @@
...
+- `target`: the element that would be considered matching if the cursor
+  eventually results in a complete match
+
+- `complexSelectorParts`: the split complex selector which the cursor is
+  attempting to match
+
+- `index`: the index into `.complexSelectorParts.compoundSelectors` of the last
+  successfully matched compound selector
+
+- `position`: the element which successfully matched the compound selector at
```
bicknellr, line 51: I forgot to update this when renaming `position` &rarr; `matchesElement`.

packages/shadydom/src/utils.js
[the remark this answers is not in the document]
sorvell: Think it's fine.

sorvell (approved):
LGTM

bicknellr:
I've been looking into the Sauce tests since they were passing before and nothing meaningful happened between then and now that should have caused them to start completely refusing to run on a handful of older browsers. I see the same thing when running the full list of browsers on sauce locally but running those same browsers individually seems to work. I tried debugging a run with the full list and found that the older browsers were being served code that wasn't correctly compiled for their feature sets, so my best guess as of now is that Sauce Connect Proxy might be caching responses _between_ different browser sessions that the single tunnel is used for, causing the older browsers later in the list to see code generated for the newer browsers earlier in the list.

Sauce Connect Proxy used to have [a `--no-proxy-caching` flag](https://wiki.saucelabs.com/pages/viewpage.action?pageId=48365781) which is now deprecated, so next I'm going to try to inject `Cache-Control: no-cache` into the responses.

bicknellr:
The test timeout probably needs to be increased; I had to restart them a few times.

PR #519, opened 2022-08-12
Fix failing tests: patch `XMLHttpRequest` if no `EventTarget`; dispatch a 'proxy' if `currentTarget` is an own value property descriptor

bicknellr:
In at least Chrome 41, events' `currentTarget` is an own property with a broken descriptor. Setting a new descriptor on one of these events makes it impossible to restore the default behavior of `currentTarget` for that event, making it useless to any later listeners. Shady DOM occasionally has to manually dispatch an event, which involves temporarily setting a new `currentTarget` descriptor to allow Shady DOM to update this property. With this PR, if the `currentTarget` property of events is broken as it is in Chrome 41, Shady DOM will create a new object with the event as its prototype, modify that new object's `currentTarget` descriptor, and pass that new object to listeners during the manual dispatch.

`EventTarget` doesn't exist in some older browsers, so objects that would normally have `EventTarget.prototype` in their prototype chain in modern browsers instead have separate copies of those properties on other prototypes. `XMLHttpRequest` is one of these objects with `EventTarget`-like properties, but these properties were unpatched until this PR.

### DIFF
--- a/packages/shadydom/src/patch-events.js
+++ b/packages/shadydom/src/patch-events.js
@@ -359,7 +359,10 @@ function fireHandlers(event, node, phase) {
   }
 }
 
-function shadyDispatchEvent(e) {
+function shadyDispatchEvent(originalEvent) {
+  const e = utils.settings.hasDescriptors
+    ? originalEvent
+    : Object.create(originalEvent);
   const path = e.composedPath();
   const retargetedPath = path.map((node) => retarget(node, path));
   const bubbles = e.bubbles;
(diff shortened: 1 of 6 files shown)


packages/shadydom/src/patch-events.js
```diff
@@ -359,7 +359,38 @@ function fireHandlers(event, node, phase) {
...
+// above process of using a proxy-like (not a `Proxy`) object for the event
+// fails because any user that attempts to use properties of the original event
+// will be using the proxy as `this`.
+//
+// Fortunately, these two issues seem to be mutually exclusive amongst the
+// browsers that we test on, so we can specifically check if `currentTarget`
+// exists on a new `Event` and switch based off of that. Safari 9 both doesn't
+// have working descriptors for `Event` properties and doesn't allow getter
+// contexts to be anything other than the original event, so
+// `utils.settings.hasDescriptors` is not a sufficient test for deciding this
```
kevinpschaaf, line 384: Hmm, so how does this not break Safari 9? Which path does it take?

bicknellr: Good catch, my comment is wrong. Safari 9 actually does have working `Event` descriptors, so it's OK to overwrite `currentTarget` on the original event. The real problem is that `utils.settings.hasDescriptors` is false in Safari 9 because it's set like this:

https://github.com/webcomponents/polyfills/blob/08c65f4203b16762b53f220f786fc293828a054a/packages/shadydom/src/utils.js#L31-L33

And Safari 9's `firstChild` descriptor is not configurable.

bicknellr: Updated.

kevinpschaaf (approved):
For posterity, because this took me a bit to grok:

On Safari 9, `currentTarget` is not an own property of `Event`, but the descriptor on the prototype is _also_ not usable (making it somewhat unique, since the unusable descriptors are all normally `ownProperty`s). So Safari 9 does go through the "save instance descriptor / finally restore descriptor" logic, but because the original descriptor is undefined, we just delete the shimmed one off the instance, and the prototype-but-not-good-accessor one is allowed to work again.

PR #520, opened 2022-09-06
Migrate the `formdata` event polyfill's tests to WTR.

bicknellr:
This PR is very similar to #509. That PR already set up WTR within the repo and this package doesn't have anything fancy going on in its tests, so this PR was just a matter of updating the wrapper in each test file.

- [ ] Try making another `.html` &rarr; `.test.html` renaming PR to reduce the diff here, which is actually very small.


bicknellr:
Looks like the tests here might also be fixed by #522.

PR #524, opened 2022-09-15
Patch the registry class only when polyfilling

### LINKED ISSUE #340: window.CustomElementRegistry is always overwritten, even when the polyfill isn't needed
### Description

This code is always run, causing `window.CustomElementRegistry` to be overwritten with the polyfilled version, even when the polyfill isn't used:
https://github.com/webcomponents/polyfills/blob/de5c30c3a51e0890bbecf6dc44b5e8364201730a/packages/custom-elements/ts_src/CustomElementRegistry.ts#L330-L331

### Example

https://jsbin.com/cotebedixo/edit?html,console,output

### Steps to reproduce

1. Add the polyfill to a page.
2. Run `customElements instanceof CustomElementRegistry` in the console.

#### Expected behavior

`customElements instanceof CustomElementRegistry` is always true.

#### Actual behavior

`customElements instanceof CustomElementRegistry` is false when the polyfill is included in a page but isn't needed.

### Version

`@webcomponents/webcomponentsjs@2.4.3`
`@webcomponents/custom-elements@1.4.1`

### Browsers affected

- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari
- [ ] IE 11

rictic:
Currently we always write/overwrite window.CustomElementsRegistry when the polyfill loads, even if we don't patch anything else, including window.customElements

If the browser has a native implementation of custom elements, this means that customElements isn't an instance of CustomElementsRegistry, so patching CustomElementsRegistry.prototype, as redefine_custom_elements does, does not effect how customElements behaves. This leads to fun and mysterious bugs!

Better to only write to CustomElementsRegistry when we're doing other patching, and otherwise leave it alone.

Fixes #340.

### DIFF
--- a/packages/custom-elements/CHANGELOG.md
+++ b/packages/custom-elements/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 <!-- ## Unreleased -->
 
+## Unreleased
+
+- Don't write to window.CustomElementsRegistry unless we're also writing to window.customElements. ([#524](https://github.com/webcomponents/polyfills/pull/524))
+
 ## [1.5.0] - 2021-08-02
 
 - Add TS externs. ([#457](https://github.com/webcomponents/polyfills/pull/457))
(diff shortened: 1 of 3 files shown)


bicknellr:
LGTM with a changelog entry, thanks!

bicknellr (approved):

PR #525, opened 2022-09-20
Update TypeScript and upstream internal changes.

bicknellr:
This PR upstreams a couple of internal CLs that added `any` in a few places to make this code compatible with a newer version of TypeScript (4.4 I think?). The PR then updates TypeScript to the latest version and makes other type changes to work with that version.


bicknellr:
Adding the "Skip Changelog" label as this PR has no user-observable changes.

packages/custom-elements/ts_src/custom-elements.ts
```diff
@@ -43,7 +43,7 @@ function installPolyfill() {
   const customElements = new CustomElementRegistry(internals);
 
   // The main document is associated with the global registry.
-  document.__CE_registry = customElements;
+  (document.__CE_registry as unknown) = customElements;
```
bicknellr, line 46: I should probably update this so that `document.__CE_registry` is defined as an instance of the polyfilled `CustomElementRegistry`, not the built-in one. The reason this can't be assigned directly is that the polyfilled `CustomElementRegistry`'s `whenDefined`'s returned promise can resolve to `undefined` in some cases when `polyfillDefineLazy` is used. Hopefully in the future we can drop that feature, since I can't find any users on [GitHub](https://github.com/search?q=polyfillLazyDefine&type=code) or [internally](http://cs/polyfillLazyDefine).

packages/custom-elements/ts_src/CustomElementInternals.ts
```diff
@@ -280,8 +280,8 @@ export default class CustomElementInternals {
       if (definition) {
         this._upgradeAnElement(element, definition);
       }
-    } catch (e) {
-      this.reportTheException(e);
+    } catch (e: unknown) {
+      this.reportTheException(e as Error);
```
bicknellr, line 284: We have a rule that prohibits `any`, so I've used `unknown` here and then cast to `Error`. `reportTheException` should probably take `unknown` instead.

bicknellr:
Hmm, I'm gonna go back to draft.

packages/formdata-event/ts_src/environment/event.ts
```diff
@@ -9,7 +9,7 @@
  * additional IP rights grant found at http://polymer.github.io/PATENTS.txt
  */
 
-export const constructor = window.Event;
+export const constructor: typeof Event = window.Event;
```
bicknellr, line 12: This probably needs a comment. Without this explicit type, TS seems to believe that `window.Event` is the wrapper from `wrappers/event.ts` at this point in time, but that wrapper is defined in terms of this variable, which causes a cycle in the type.

packages/custom-elements/ts_src/CustomElementInternals.ts
bicknellr: `reportTheException` now takes unknown.

packages/custom-elements/ts_src/custom-elements.ts
bicknellr: `__CE_registry` is now defined using the polyfill type.

packages/formdata-event/ts_src/environment/event.ts
bicknellr: Added.

bicknellr:
Internal tests of the new polyfills including this CL passed so I think this is ready for review now.

rictic (approved):

bicknellr:
Thanks for the review!

Issue #459, opened 2021-07-19
[scoped-custom-element-registry]: Support for JSDOM (Jest)


riovir:
## Description

Please consider altering the polyfill to keep JSDOM's global `customElements` registry intact. This particular instance also includes some _internal bookkeeping the headless browser uses. Swapping it out completely breaks JSDOM down.

_However_ by swapping it out with a `Proxy`-ed object that delegates to the polyfill's instance where it can, and allows fallthrough to the original instance seems to also add support to JSDOM. 

## Motivation

So why consider a "not a real browser" as some would put it? Currently, a large percentage of applications written with traditional frameworks (React or Vue) use Jest including JSDOM (v16+). Currently, web components are often sold to management as a future-proofing investment. Now if we take an organization adopting, say [Lion Web Components](https://lion-web.netlify.app/), suddenly all their Jest user teams' tests start breaking after [@open-wc/scoped-elements](https://www.npmjs.com/package/@open-wc/scoped-elements) has updated to use this polyfill. 

## Note

As far as I managed to test it, the following [change to the polyfill](https://github.com/webcomponents/polyfills/blob/master/packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js#L400-L405) would solve the problem for JSDOM. I'm happy to send a PR if needed:

```js
	// Install global registry
	const globalRegistry = new CustomElementRegistry();
	Object.defineProperty(window, 'customElements', {
		value: new Proxy(window.customElements, {
			get: (original, prop) => prop in globalRegistry ?
				globalRegistry[prop].bind(globalRegistry) :
				original[prop],
		}),
		configurable: true,
		writable: true,
	});
```

bicknellr:
I was talking with @kevinpschaaf yesterday and he discovered this:

https://github.com/jsdom/jsdom/blob/04f6c13f4a4d387c7fc979b8f62c6f68d8a0c639/lib/jsdom/living/helpers/html-constructor.js#L13-L18

It looks like jsdom is grabbing the main registry by reading the `.customElements` property of the global, which is not something the browser would do. I'm going to try and make a PR for jsdom.

PR #530, opened 2022-10-24
[scoped-custom-element-registry] Add basic `jsdom` tests

bicknellr:
jsdom currently has a bug where it reads the global custom elements registry by looking through the sandbox-facing `customElements` property of its global object. This means that scripts running in that context can break jsdom's custom elements implementation by overwriting or deleting the `customElements` global.

I'm making a PR to fix jsdom upstream. Once `jsdom` and `jest-environment-jsdom` are released with updated versions, these tests should pass.

Closes <https://github.com/webcomponents/polyfills/issues/459>.

### DIFF
--- a/packages/scoped-custom-element-registry/jest.config.js
+++ b/packages/scoped-custom-element-registry/jest.config.js
@@ -1,0 +1,4 @@
+module.exports = {
+  testEnvironment: "jsdom",
+  rootDir: "test/jsdom",
+};
--- a/packages/scoped-custom-element-registry/package-lock.json
+++ b/packages/scoped-custom-element-registry/package-lock.json
@@ -1,6458 +1,13478 @@
 {
-  "name": "@webcomponents/scoped-custom-element-registry",
-  "version": "0.0.5",
-  "lockfileVersion": 2,
-  "requires": true,
-  "packages": {
-    "": {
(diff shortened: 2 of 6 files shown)


bicknellr:
[jsdom 20.0.2](https://github.com/jsdom/jsdom/commit/e285763ebf46bbc9c883a519c9a18231f5ede9d8) contains the PR with the fix. `jest-environment-jsdom` depends on `jsdom@^20.0.0` so its not guaranteed to use `20.0.2`. Adding `jsdom@^20.0.2` as a dev dependency here would satisfy node's deduping during installation of `jest-environment-jsdom`, but I don't think it's a great idea to rely on that behavior so I'll keep this open until it depends on at least `20.0.2`.

bicknellr:
> Adding `jsdom@^20.0.2` as a dev dependency here would satisfy node's deduping during installation of `jest-environment-jsdom`

I'm trying this out here: https://github.com/webcomponents/polyfills/commit/39e3f06c7b2be69bbe9b0b1c3220480a81409095

bicknellr:
Except for the formatting, the tests passed on that commit.

bicknellr:
Added "Skip Changelog" label because there are no user-facing changes.

Issue #533, opened 2022-11-20
[scoped-custom-element-registry] toggleAttribute polyfill does not retain force argument


jmcgavin:
### Description

When calling `toggleAttribute`, [the polyfill](https://github.com/webcomponents/polyfills/blob/master/packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js#L349) does not retain the [`force` argument](https://developer.mozilla.org/en-US/docs/Web/API/Element/toggleAttribute#syntax).

This means that when I call `toggleAttribute('my-attribute', false)`, the attribute is applied when it should not be.

### Example

```js
// Include the scoped-custom-element-registry polyfill

this.renderRoot.querySelector('my-custom-element').toggleAttribute('my-attribute', false)
```

### Steps to reproduce

1. Call `toggleAttribute` on an element
2. Provide `'my-attribute'` as the first argument
3. Provide `false` as the second argument

#### Expected behavior

`my-attribute` should not be applied to the element.

#### Actual behavior

`my-attribute` is applied to the element.

### Version

@webcomponents/scoped-custom-element-registry@0.0.6

### Browsers affected

- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari
- [x] IE 11

bicknellr:
The fix for this is available in [`v0.0.8`](https://www.npmjs.com/package/@webcomponents/scoped-custom-element-registry/v/0.0.8).

PR #534, opened 2022-11-20
fix: toggleAttribute polyfill should retain force argument

jmcgavin:
Addresses an issue where `toggleAttribute` would not retain `force`, the second positional argument passed to it.

Fixes https://github.com/webcomponents/polyfills/issues/533

### DIFF
--- a/packages/scoped-custom-element-registry/test/Element.test.html.js
+++ b/packages/scoped-custom-element-registry/test/Element.test.html.js
@@ -146,7 +146,7 @@ describe('Element', () => {
 
       expect($el.hasAttribute('foo')).to.be.false;
 
-      $el.setAttribute('foo');
+      $el.setAttribute('foo', '');
       $el.toggleAttribute('foo', true);
 
       expect($el.hasAttribute('foo')).to.be.true;
--- a/packages/scoped-custom-element-registry/test/Element.test.html.js
+++ b/packages/scoped-custom-element-registry/test/Element.test.html.js
@@ -4,7 +4,7 @@ import {
   getTestElement,
   getObservedAttributesTestElement,
   getShadowRoot,
-  getHTML
+  getHTML,
 } from './utils.js';
 
 describe('Element', () => {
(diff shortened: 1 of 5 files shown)


u-b-i-p-i-x-e-l-s:
Hey @aomarks and @bicknellr 👋 

Could you please review? Thanks!!

bicknellr:
Hi, thanks for sending this! Could you add a test or two? It doesn't seem like we have any for the attribute functions on `Element` yet, but `Element.test.html.js` seems like it would be the right place to put it.

packages/scoped-custom-element-registry/test/Element.test.html.js
```diff
@@ -104,4 +109,41 @@ describe('Element', () => {
       expect($el.firstElementChild).to.not.be.instanceof(CustomElementClass);
     });
   });
+
+  describe('attributes', () => {
+    it('should call setAttribute', () => {
+      const {tagName, CustomElementClass} = getObservedAttributesTestElement(['foo']);
+      customElements.define(tagName, CustomElementClass);
+      const $el = getHTML('<div></div>');
+      $el.innerHTML = `<${tagName}></${tagName}>`;
```
bicknellr, line 118: I think this could be simplified to:

```suggestion
      const $el = document.createElement(tagName);
```

here and in the last test.

packages/scoped-custom-element-registry/test/Element.test.html.js
```diff
@@ -104,4 +109,41 @@ describe('Element', () => {
...
+
+      $el.firstElementChild.setAttribute('foo', 'bar');
+
+      expect($el.firstElementChild.getAttribute('foo')).to.equal('bar');
```
bicknellr, line 122: And then the `.firstElementChild` lookups could be removed down below here and in the other tests:

```suggestion
      $el.setAttribute('foo', 'bar');

      expect($el.getAttribute('foo')).to.equal('bar');
```

packages/scoped-custom-element-registry/test/Element.test.html.js
```diff
@@ -104,4 +109,41 @@ describe('Element', () => {
...
+      const {tagName, CustomElementClass} = getObservedAttributesTestElement(['foo']);
+      customElements.define(tagName, CustomElementClass);
+      const $el = getHTML('<div></div>');
+      $el.innerHTML = `<${tagName}></${tagName}>`;
+
+      $el.firstElementChild.toggleAttribute('foo', false);
+
+      expect($el.firstElementChild.hasAttribute('foo')).to.be.false;
+
+      $el.firstElementChild.toggleAttribute('foo', true);
```
bicknellr, line 144: I think it would be good to explicitly set the attribute to the empty string before this `.toggleAttribute('foo', true)` to confirm that this call leaves the attribute there if it was already set.

packages/scoped-custom-element-registry/test/utils.js
```diff
@@ -30,6 +30,20 @@ export const getTestElement = () => ({
   CustomElementClass: class extends HTMLElement {},
 });
 
+/**
+ * 
+ * @param {array<string>} observedAttributeNames the names of the attributes you want to observe
+ * @returns {{Klass: typeof HTMLElement, tagName: string}}
```
bicknellr, line 36:
```suggestion
 * @param {Array<string>} observedAttributeNames the names of the attributes you want to observe
 * @returns {{CustomElementClass: typeof HTMLElement, tagName: string}}
```

packages/scoped-custom-element-registry/test/Element.test.html.js
```diff
@@ -104,4 +109,41 @@ describe('Element', () => {
...
+      const {tagName, CustomElementClass} = getObservedAttributesTestElement(['foo']);
+      customElements.define(tagName, CustomElementClass);
+      const $el = getHTML('<div></div>');
+      $el.innerHTML = `<${tagName} foo></${tagName}>`;
+
+      $el.firstElementChild.removeAttribute('foo');
+
+      expect($el.firstElementChild.hasAttribute('foo')).to.be.false;
+    });
+    it('should call toggleAttribute', () => {
```
bicknellr, line 134:
```suggestion

    it('should call toggleAttribute', () => {
```

packages/scoped-custom-element-registry/test/Element.test.html.js
```diff
@@ -104,4 +109,41 @@ describe('Element', () => {
...
+      const {tagName, CustomElementClass} = getObservedAttributesTestElement(['foo']);
+      customElements.define(tagName, CustomElementClass);
+      const $el = getHTML('<div></div>');
+      $el.innerHTML = `<${tagName}></${tagName}>`;
+
+      $el.firstElementChild.setAttribute('foo', 'bar');
+
+      expect($el.firstElementChild.getAttribute('foo')).to.equal('bar');
+    });
+    it('should call removeAttribute', () => {
```
bicknellr, line 124:
```suggestion

    it('should call removeAttribute', () => {
```

jpzwarte:
@jmcgavin any chance you could incorporate the review comments? Do you mind if somebody takes over?

jmcgavin:
@jpzwarte Apologies for the delay! I've addressed the comments that were left

jmcgavin:
@bicknellr Thank you for the review and my apologies for the tardiness of my response. I've addressed your comments. Let me know if it requires additional changes or tests

bicknellr:
Thanks for the updates!

I looked into the failing test and it's seems like an unrelated bug that was exposed by a new `getHTML` call. This line is unconditionally calling `attributeChangedCallback` even though it might be `undefined`:

https://github.com/webcomponents/polyfills/blob/0ab5679e40e1363cceb8e7f4d8279d3bdfebce1f/packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js#L393

The whole outer iteration over `observedAttributes` could be nested in an `if (definition.attributeChangedCallback) { ... }` to work around this:

https://github.com/webcomponents/polyfills/blob/0ab5679e40e1363cceb8e7f4d8279d3bdfebce1f/packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js#L391-L400

Also, could you add a change log entry? ([example here](https://github.com/webcomponents/polyfills/commit/7dc9123fdd371b5f5dce9b1b3d3dba0d4a604521#diff-f8814c39c58e56683d9b22c090880894d2ce76a29ae7e99862d864e885dbb148))

bicknellr (commented):
One last thing, thanks!

bicknellr (approved):
Thanks!

PR #537, opened 2023-01-04
fix(scoped-custom-elements-registry): return value from HTMLElement p…

calebdwilliams:
…rototype methods in polyfill

This fixes an issue where the scoped custom element registry polyfill fails to return the value of a polyfilled internals method.

### DIFF
--- a/packages/scoped-custom-element-registry/CHANGELOG.md
+++ b/packages/scoped-custom-element-registry/CHANGELOG.md
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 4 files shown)


michaelwarren1106:
this is a great change. In components where the scoped elements registry is used, if the value of those functions isn't returned, other mixins/polyfills like the `element-internals-polyfill` which also polyfills ElementInternals aspects won't be able to function properly.

justinfagnani:
@calebdwilliams thanks! Any chance of adding a test for this?

justinfagnani:
@calebdwilliams you need to update the CHANGELOG for scoped-custom-element-registry

calebdwilliams:
Cool. I’ll take a crack at those tonight.

calebdwilliams:
Updated the PR with a CHANGELOG entry and added a test.

justinfagnani (approved):
LGTM, but @kevinpschaaf should approve

kevinpschaaf (approved):
LGTM

PR #541, opened 2023-03-23
Add Element.toggleAttribute() polyfill and support in Shady DOM

justinfagnani:


packages/shadydom/src/wrapper.js
```diff
@@ -87,6 +87,10 @@ class Wrapper {
     this.node[utils.SHADY_PREFIX + 'removeAttribute'](name);
   }
 
+  toggleAttribute(name, force) {
+    this.node[utils.SHADY_PREFIX + 'toggleAttribute'](name, force);
```
bicknellr, line 91:
```suggestion
    return this.node[utils.SHADY_PREFIX + 'toggleAttribute'](name, force);
```

packages/formdata-event/ts_src/environment/element.ts
```diff
@@ -18,4 +18,5 @@ export const methods = {
   hasAttribute: prototype.hasAttribute,
   removeAttribute: prototype.removeAttribute,
   setAttribute: prototype.setAttribute,
+  toggleAttribute: prototype.toggleAttribute,
```
bicknellr, line 21: I don't think this is used anywhere but also, meh, I'd not worry about adding `toggleAttribute` support to the the formdata event polyfill: it was never released and might already be obsolete, IIRC.

packages/shadydom/src/patches/Element.js
bicknellr, line 1: cc @sorvell, since I don't know these details of Shady DOM.

packages/shadydom/src/patches/Element.js
```diff
@@ -120,6 +120,23 @@ export const ElementPatches = utils.getOwnPropertyDescriptors({
...
   },
+
+  /**
+   * @this {Element}
+   * @param {string} attr
+   * @param {boolean | undefined} force
+   */
+  toggleAttribute(attr, force) {
+    if (this.ownerDocument !== doc) {
+      this[utils.NATIVE_PREFIX + 'toggleAttribute'](attr, force);
```
bicknellr, line 131:
```suggestion
      return this[utils.NATIVE_PREFIX + 'toggleAttribute'](attr, force);
```
(and on the others below)

packages/shadydom/src/patches/Element.js
```diff
@@ -120,6 +120,23 @@ export const ElementPatches = utils.getOwnPropertyDescriptors({
...
+    } else if (!scopeClassAttribute(this, attr, '')) {
+      this[utils.NATIVE_PREFIX + 'toggleAttribute'](attr, force);
+      distributeAttributeChange(this, attr);
+    } else if (this.getAttribute(attr) === '' && !force) {
+      // ensure that "class" attribute is fully removed if ShadyCSS does not keep scoping
+      this[utils.NATIVE_PREFIX + 'toggleAttribute'](attr, force);
+    }
```
bicknellr, line 138: Is one of these cases guaranteed to happen?

packages/formdata-event/ts_src/environment/element.ts
justinfagnani: I was just adding it everywhere there were attribute methods... seem ok just in case this did get used in the future and toggleAttribute didn't work. If we remove the whole package, that's fine too.

justinfagnani:
Added some tests. I think this is _mostly_ in the right shape now, except for actually adding a `toggleAttribute()` polyfill for IE. The tests at least cover `toggleAttribute()` patching behavior in browsers that have it natively.

justinfagnani:
Ready for review!

packages/shadydom/src/patches/Element.js
justinfagnani: cc @sorvell

packages/tests/webcomponentsjs_/toggle-attribute.html
```diff
@@ -0,0 +1,51 @@
...
+Copyright (c) 2020 The Polymer Project Authors. All rights reserved.
+This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+Code distributed by Google as part of the polymer project is also
+subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+-->
+<html>
+  <head>
+    <title>Element getAttributeNames</title>
```
sorvell, line 13:
```suggestion
    <title>Element toggleAttribute</title>
```

packages/formdata-event/ts_src/environment/element.ts
bicknellr: Adding it here doesn't do anything on its own though. This file is kind of like the `Native.ts` in the custom elements polyfill: it just captures the native stuff at the time the polyfill is installed so that it can be used by something else later.

https://github.com/webcomponents/polyfills/blob/4dc188e9afd53a3465cedcaefe0aea7da80b076f/packages/formdata-event/ts_src/environment/element.ts#L18

https://github.com/webcomponents/polyfills/blob/4dc188e9afd53a3465cedcaefe0aea7da80b076f/packages/formdata-event/ts_src/environment_api/element.ts#L12-L16

https://github.com/webcomponents/polyfills/blob/4dc188e9afd53a3465cedcaefe0aea7da80b076f/packages/formdata-event/ts_src/dispatch_formdata_for_submission.ts#L25

https://github.com/webcomponents/polyfills/blob/4dc188e9afd53a3465cedcaefe0aea7da80b076f/packages/formdata-event/ts_src/dispatch_formdata_for_submission.ts#L106

packages/shadydom/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 <!-- ## Unreleased -->
```
bicknellr, line 8:
```suggestion
## Unreleased
```

packages/shadydom/src/patches/Element.js
```diff
@@ -120,6 +120,23 @@ export const ElementPatches = utils.getOwnPropertyDescriptors({
...
+   * @this {Element}
+   * @param {string} attr
+   * @param {boolean | undefined} force
+   */
+  toggleAttribute(attr, force) {
+    if (this.ownerDocument !== doc) {
+      return this[utils.NATIVE_PREFIX + 'toggleAttribute'](attr, force);
+    } else if (!scopeClassAttribute(this, attr, '')) {
+      this[utils.NATIVE_PREFIX + 'toggleAttribute'](attr, force);
+      distributeAttributeChange(this, attr);
```
bicknellr, line 134:
```suggestion
      const result = this[utils.NATIVE_PREFIX + 'toggleAttribute'](attr, force);
      distributeAttributeChange(this, attr);
      return result;
```

packages/shadydom/src/patches/Element.js
```diff
@@ -120,6 +120,23 @@ export const ElementPatches = utils.getOwnPropertyDescriptors({
...
+    } else if (this.getAttribute(attr) === '' && !force) {
+      // ensure that "class" attribute is fully removed if ShadyCSS does not keep scoping
+      this[utils.NATIVE_PREFIX + 'toggleAttribute'](attr, force);
```
bicknellr, line 137:
```suggestion
      return this[utils.NATIVE_PREFIX + 'toggleAttribute'](attr, force);
```

packages/tests/custom-elements/html/Element/toggleAttribute.test.html
```diff
@@ -0,0 +1,221 @@
...
+          });
+
+          test('Toggling an unobserved attribute removes the attribute but does not trigger a callback.', () => {
+            const element = document.createElement(localName1);
+
+            assert.equal(element.attrCallbackArgs.length, 0);
+
+            element.setAttribute('attr', 'abc');
+
+            assert.equal(element.attrCallbackArgs.length, 0);
```
bicknellr, line 123:
```suggestion
            assert.equal(element.getAttribute('attr'), 'abc');
            assert.equal(element.attrCallbackArgs.length, 0);
```
Maybe this is overkill but I kinda like confirming the state right before doing the thing in question when it relies on earlier stuff in the tests.

packages/tests/custom-elements/html/Element/toggleAttribute.test.html
```diff
@@ -0,0 +1,221 @@
...
+          });
+
+          test('Toggling (force: true) an unobserved attribute does not change the attribute and does not trigger a callback.', () => {
+            const element = document.createElement(localName1);
+
+            assert.equal(element.attrCallbackArgs.length, 0);
+
+            element.setAttribute('attr', 'abc');
+
+            assert.equal(element.attrCallbackArgs.length, 0);
```
bicknellr, line 138:
```suggestion
            assert.equal(element.getAttribute('attr'), 'abc');
            assert.equal(element.attrCallbackArgs.length, 0);
```

packages/tests/custom-elements/html/Element/toggleAttribute.test.html
```diff
@@ -0,0 +1,221 @@
...
+          });
+
+          test('Toggling (force: false) an unobserved attribute removes the attribute but does not trigger a callback.', () => {
+            const element = document.createElement(localName1);
+
+            assert.equal(element.attrCallbackArgs.length, 0);
+
+            element.setAttribute('attr', 'abc');
+
+            assert.equal(element.attrCallbackArgs.length, 0);
```
bicknellr, line 153:
```suggestion
            assert.equal(element.getAttribute('attr'), 'abc');
            assert.equal(element.attrCallbackArgs.length, 0);
```

packages/tests/custom-elements/html/Element/toggleAttribute.test.html
```diff
@@ -0,0 +1,221 @@
...
+          });
+
+          test('Toggling an observed attribute removes the attribute and triggers a callback.', () => {
+            const element = document.createElement(localName2);
+
+            assert.equal(element.attrCallbackArgs.length, 0);
+
+            element.setAttribute('attr', 'abc');
+
+            assert.equal(element.attrCallbackArgs.length, 1);
```
bicknellr, line 168:
```suggestion
            assert.equal(element.getAttribute('attr'), 'abc');
            assert.equal(element.attrCallbackArgs.length, 1);
```

packages/tests/custom-elements/html/Element/toggleAttribute.test.html
```diff
@@ -0,0 +1,221 @@
...
+          });
+
+          test('Toggling (force: true) an observed attribute does not change the attribute and does not trigger a callback.', () => {
+            const element = document.createElement(localName2);
+
+            assert.equal(element.attrCallbackArgs.length, 0);
+
+            element.setAttribute('attr', 'abc');
+
+            assert.equal(element.attrCallbackArgs.length, 1);
```
bicknellr, line 189:
```suggestion
            assert.equal(element.getAttribute('attr'), 'abc');
            assert.equal(element.attrCallbackArgs.length, 1);
```

packages/tests/custom-elements/html/Element/toggleAttribute.test.html
```diff
@@ -0,0 +1,221 @@
...
+          });
+
+          test('Toggling (force: false) an observed attribute removes the attribute and triggers a callback.', () => {
+            const element = document.createElement(localName2);
+
+            assert.equal(element.attrCallbackArgs.length, 0);
+
+            element.setAttribute('attr', 'abc');
+
+            assert.equal(element.attrCallbackArgs.length, 1);
```
bicknellr, line 204:
```suggestion
            assert.equal(element.getAttribute('attr'), 'abc');
            assert.equal(element.attrCallbackArgs.length, 1);
```

packages/shadydom/src/patches/Element.js
sorvell: Yeah, the code here is pretty messy unfortunately, and the test covereage is really poor. 

I would suggest just calling `this[utils.SHADY_PREFIX + 'set|removeAttribute'](...);` with the code you have in platform. (What is here is incorrect, at the least, because shadycss will always try to set the attribute.)

The shadydom tests are really spotty and messy but we should add something for this. Perhaps it could be dumped in here and modeled after a smaller subset of this test: https://github.com/webcomponents/polyfills/blob/4756d228c16a0f53a739846edeaebd740bb9dc09/packages/tests/shadydom/shady.html#L527.

sorvell (changes requested):

packages/custom-elements/ts_src/Patch/Element.ts
```diff
@@ -179,6 +179,31 @@ export default function (internals: CustomElementInternals) {
     }
   };
 
+  Element.prototype.toggleAttribute = function (
```
bicknellr, line 182: It looks like the tests are failing because they're being run unconditionally in browsers that don't have native support for `toggleAttribute`, which also reminded me that we shouldn't write `toggleAttribute` if a native one didn't exist before. If you wrap this block in a check for `Native.Element_toggleAttribute`

https://github.com/webcomponents/polyfills/blob/4dc188e9afd53a3465cedcaefe0aea7da80b076f/packages/custom-elements/ts_src/Patch/Element.ts#L21-L22

and then skip the root suites in the new tests

https://github.com/webcomponents/polyfills/blob/4dc188e9afd53a3465cedcaefe0aea7da80b076f/packages/tests/custom-elements/html/Interface/ChildNode/index.test.html#L49-L58

that should fix these.

bicknellr (changes requested):

packages/custom-elements/ts_src/Patch/Element.ts
justinfagnani: So we're not loading the platform polyfill for these test?

packages/shadydom/src/patches/Element.js
justinfagnani: I think that implementing this in polyfill-style for toggleAttribute on top of setAttribute/removeAttribute also has a chance of being buggy, for which we need test coverage, and that if you distribute the cases in toggleAttribute through these set/removeAttribute methods you get to the logic here.

The main case is the middle one: `if (!scopeClassAttribute(this, attr, ''))` which happens if `attr` is _not_ class, and the same condition is used in both setAttribute and removeAttribute. The last case is only implemented in in removeAttribute, and so replicated here. The cases that fall though here, by inspection, also fall though in both setAttribute and removeAttribute.

If we think there are cases that might be missed and need to be covered by testing, I would like to add them to the tests regardless of implementation. I added versions of all tests I could find that already exist against set/remove and those pass.

sorvell: OK, that makes sense to me. Thanks for adding the extra tests.

packages/shadydom/src/patches/Element.js
```diff
@@ -130,11 +130,12 @@ export const ElementPatches = utils.getOwnPropertyDescriptors({
     if (this.ownerDocument !== doc) {
       return this[utils.NATIVE_PREFIX + 'toggleAttribute'](attr, force);
     } else if (!scopeClassAttribute(this, attr, '')) {
-      this[utils.NATIVE_PREFIX + 'toggleAttribute'](attr, force);
+      const result = this[utils.NATIVE_PREFIX + 'toggleAttribute'](attr, force);
       distributeAttributeChange(this, attr);
+      return result;
     } else if (this.getAttribute(attr) === '' && !force) {
```
sorvell, line 136: Can you explain why the check for `!force` is here?

packages/webcomponentsjs/ts_src/platform/toggle-attribute.ts
```diff
@@ -0,0 +1,41 @@
...
+http://polymer.github.io/LICENSE.txt The complete set of authors may be found at
+http://polymer.github.io/AUTHORS.txt The complete set of contributors may be
+found at http://polymer.github.io/CONTRIBUTORS.txt Code distributed by Google as
+part of the polymer project is also subject to an additional IP rights grant
+found at http://polymer.github.io/PATENTS.txt
+*/
+
+export {};
+
+const Element_prototype = Element.prototype;
```
bicknellr, line 14:
```suggestion
const Element_prototype = Element.prototype;
const nativeHasAttribute = Element.prototype.hasAttribute;
const nativeSetAttribute = Element.prototype.setAttribute;
const nativeRemoveAttribute = Element.prototype.removeAttribute;
```

These need to be saved to avoid calling back into other polyfills that were applied after this one.

packages/webcomponentsjs/ts_src/platform/toggle-attribute.ts
```diff
@@ -0,0 +1,41 @@
...
+
+if (!Element_prototype.hasOwnProperty('toggleAttribute')) {
+  Element_prototype.toggleAttribute = function toggleAttribute(
+    this: Element,
+    name: string,
+    force?: boolean
+  ): boolean {
+    if (force === undefined) {
+      if (this.hasAttribute(name)) {
+        this.removeAttribute(name);
```
bicknellr, line 24:
```suggestion
      if (nativeHasAttribute.call(this, name)) {
        nativeRemoveAttribute.call(this, name);
```

packages/webcomponentsjs/ts_src/platform/toggle-attribute.ts
```diff
@@ -0,0 +1,41 @@
...
+        return false;
+      } else {
+        this.setAttribute(name, '');
```
bicknellr, line 27:
```suggestion
        nativeSetAttribute.call(this, name, '');
```

packages/webcomponentsjs/ts_src/platform/toggle-attribute.ts
```diff
@@ -0,0 +1,41 @@
...
+        return true;
+      }
+    }
+    if (force) {
+      if (!this.hasAttribute(name)) {
+        this.setAttribute(name, '');
```
bicknellr, line 33:
```suggestion
      if (!nativeHasAttribute.call(this, name)) {
        nativeSetAttribute.call(this, name, '');
```

packages/webcomponentsjs/ts_src/platform/toggle-attribute.ts
```diff
@@ -0,0 +1,41 @@
...
+      }
+      return true;
+    }
+    // force is falsey
+    this.removeAttribute(name);
```
bicknellr, line 38:
```suggestion
    nativeRemoveAttribute.call(this, name);
```

bicknellr (changes requested):

packages/shadydom/src/patches/Element.js
justinfagnani: Added explanation in comment and updated the comment in removeAttribute()

packages/formdata-event/ts_src/environment/element.ts
justinfagnani: removed

bicknellr (approved):
I guess one thing we might want to test is the return values from `toggleAttribute`, otherwise LGTM.

PR #543, opened 2023-03-30
Prepare releases for packages with toggleAttribute feature

justinfagnani:
Prepares releases for:

- @webcomponents/custom-elements@1.6.0
- @webcomponents/shadydom@1.11.0
- @webcomponents/webcomponentsjs@2.8.0

Other packages with just their semver-compatible deps updated in https://github.com/webcomponents/polyfills/pull/542 don't need releases because those only updated their package locks.


rictic (approved):

bicknellr (changes requested):
Let me handle the version bumping etc. as part of the release. I need to get changesets set up so we don't have to do this manually any more. :/

Issue #419, opened 2021-01-26
[scoped-custom-element-registry] Convert to TypeScript


kevinpschaaf:
Task: The [scoped-custom-element-registry](https://github.com/webcomponents/polyfills/tree/master/packages/scoped-custom-element-registry) polyfill was prototyped in JavaScript, but should be upgraded to TypeScript to match the conventions of the monorepo.

PR #551, opened 2023-08-21
Use wireit in scoped-custom-element-registry

justinfagnani:
This is the minimal change to use Wireit to run the existing Gulp build script, but get builds to run as a dependency of tests. Next I would like to replace Gulp with Wireit and add TypeScript.

I'm doing this as part of https://github.com/webcomponents/polyfills/issues/419 working towards fixing some other issues.


rictic (approved):

rictic:
Changelog

jessevanassen:
FYI, by running the tests through wireit, it now prevents them from running in watch mode with `npm run test -- --watch` because the terminal isn't a TTY any more.
As a workaround, running the tests directly with `npx wtr --watch` works.

PR #555, opened 2023-08-21
Call Closure directly from Wireit

justinfagnani:
I diffed the output file pre and post change and it is byte-for-byte identical.

The source map is different because it's generated directly by Closure now instead of gulp-sourcemaps. It should mostly be JSON key order and minification differences. From what I can tell visually, it appears largely the same.


rictic (approved):
Nice!

PR #557, opened 2023-08-22
[scoped-custom-element-registry] Toggle attribute should only trigger attributechangedcallback on change

### LINKED ISSUE #556: [scoped-custom-element-registry] toggleAttribute should only trigger attributeChangedCallback when the attribute is actually toggled
### Description
According to the [`toggleAttribute(qualifiedName, force)` spec](https://dom.spec.whatwg.org/#dom-element-toggleattribute), toggling an attribute should only result in a change if the attribute actually changes. In other words, if the attribute is already present and force is `true` or if the attribute is not present and force is `false`, the attribute shouldn't change.

When the polyfill is loaded, the attribute is always changed, resulting in the `attributeChangedCallback` firing too frequently.

This is also inconsistent with how it works out of the box in the latest versions of Chrome, Firefox and Edge.

It is especially problematic, because it breaks the spec _for all web components on the page_, not just components that use the scoped registry.
In our case, it lead to an unrelated piece of code behaving differently when the polyfill was loaded elsewhere.

### Example
```html
<!DOCTYPE html>
<html lang="en">
	<head>
		<meta charset="UTF-8">
		<script src="https://unpkg.com/@webcomponents/scoped-custom-element-registry@0.0.9/src/scoped-custom-element-registry.js"></script>
	</head>
	<body>
		<script type="text/javascript">
			customElements.define('my-element', class MyElement extends HTMLElement {
				static get observedAttributes() {
					return ['param'];
				}

				attributeChangedCallback(name, oldValue, newValue) {
					console.log('attributeChangedCallback', { name, oldValue, newValue });
				}
			});


			const element = document.createElement('my-element');
			document.body.append(element);

			// This should log: first setting of param
			element.setAttribute('param', '');

			// This should log: removing the param
			element.toggleAttribute('param', false);

			// This shouldn't log: parameter is not present
			element.toggleAttribute('param', false);

			// This should log: adding the parameter 
			element.toggleAttribute('param', true);

			// This shouldn't log: parameter is already present
			element.toggleAttribute('param', true);
		</script>
	</body>
</html>
```

- When the polyfill is not included, the `attributeChangedCallback` fires 3 times, resulting in 3 log statements.
- When the polyfill is included, the `attributeChangedCallback` fires 5 times, resulting in 5 log statements.

#### Expected behavior
`attributeChangedCallback` is only called when the attribute is actually toggled

#### Actual behavior
`attributeChangedCallback` is also called when the attribute presence remains the same

### Version
0.0.9

### Browsers affected

- [x] Chrome
- [x] Firefox
- [x] Edge
- [ ] Safari
- [ ] IE 11

(These are the only browsers I can test)

jessevanassen:
Fixes #556.

I fixed this by only triggering the attributeChangedCallback if the attribute's value was actually different. I've also added a test to assert the behavior.

There were no tests to assert the behavior of `attributeChangedCallback` in combination with `setAttribute`, `removeAttribute` and `toggleAttribute`, so I've added additional tests for this as well.


justinfagnani:
Looks good, but there's a formatting error on the changelog. I'm going to see if I can push to your branch to fix.

*wow, that's a long branch name! my terminal's having trouble with it.

justinfagnani:
ah, I don't have permissions to push.

justinfagnani (approved):
Thanks for the issue, fix, and tests!

You just need to rebase or merge master and fix the formatting on the changelog (`npm run format` from the repo root should do it)

jessevanassen:
Thanks for the quick response and the pointers to fix it @justinfagnani!
Apologies for the formatting issue in the changelog. Fixed it, and rebased the branch so it's up to date with master.

PR #561, opened 2023-08-22
[scoped-custom-element-registry] Port to TypeScript

justinfagnani:
Fixes https://github.com/webcomponents/polyfills/issues/419

For ease of review and checking that the output was reasonably close to the previous Closure output, I made a first commit that is almost only type-changes. I manually diffed the output and it's most variable rename differences, and very close in byte size. Then I added a commit which cleans up some of the list and type checker warnings I disabled in the first. Further cleanups can happen later.

### DIFF
--- a/packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
+++ b/packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
@@ -99,7 +99,7 @@ if (!ShadowRoot.prototype.createElement) {
     readonly promise: Promise<T>;
     readonly resolve: (val: T) => void;
     constructor() {
-      let resolve;
+      let resolve: (val: T) => void;
       this.promise = new Promise<T>((r) => {
         resolve = r;
       });
@@ -210,7 +210,7 @@ if (!ShadowRoot.prototype.createElement) {
       return this._definitionsByTag.get(tagName);
     }
 
-    whenDefined(tagName: string) {
+    ['whenDefined'](tagName: string) {
       const definition = this._getDefinition(tagName);
       if (definition !== undefined) {
         return Promise.resolve(definition.elementClass);
--- a/packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
+++ b/packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
@@ -52,8 +52,8 @@ interface CustomElementRegistry {
 interface CustomElementDefinition {
   elementClass: CustomElementConstructor;
   /**
-   * We hold onto the versions of callbacks at registration time, because that's the
-   * specc'd behavior.
+   * We hold onto the versions of callbacks at registration time, because
+   * that's the specc'd behavior.
    */
   connectedCallback?: CustomHTMLElement['connectedCallback'];
   disconnectedCallback?: CustomHTMLElement['disconnectedCallback'];
@@ -65,6 +65,11 @@ interface CustomElementDefinition {
   formResetCallback?: CustomHTMLElement['formResetCallback'];
   formStateRestoreCallback?: CustomHTMLElement['formStateRestoreCallback'];
   observedAttributes: Set<string>;
+  /**
+   * The class that's registered on the global custom element registry for this
+   * element definition. Only present if this definition is registered on the
+   * global registry, though all definitions do have a standin.
+   */
   standInClass?: CustomElementConstructor;
 }
 
--- a/packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
+++ b/packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
@@ -44,6 +44,10 @@ interface CustomElementRegistry {
 
 interface CustomElementDefinition {
   elementClass: CustomElementConstructor;
+  /**
+   * We hold onto the versions of callbacks at registration time, because that's the
+   * specc'd behavior.
+   */
   connectedCallback?: CustomHTMLElement['connectedCallback'];
   disconnectedCallback?: CustomHTMLElement['disconnectedCallback'];
   adoptedCallback?: CustomHTMLElement['adoptedCallback'];
@@ -107,16 +111,19 @@ if (!ShadowRoot.prototype.createElement) {
   // register stand-in elements that can delegate out to CE classes registered
   // in scoped registries
   class ShimmedCustomElementsRegistry implements CustomElementRegistry {
-    private _definitionsByTag = new Map<string, CustomElementDefinition>();
-    private _definitionsByClass = new Map<
+    private readonly _definitionsByTag = new Map<
+      string,
+      CustomElementDefinition
+    >();
+    private readonly _definitionsByClass = new Map<
       CustomElementConstructor,
       CustomElementDefinition
     >();
-    private _whenDefinedPromises = new Map<
+    private readonly _whenDefinedPromises = new Map<
       string,
       AsyncInfo<CustomElementConstructor>
     >();
-    private _awaitingUpgrade = new Map<string, Set<HTMLElement>>();
+    private readonly _awaitingUpgrade = new Map<string, Set<HTMLElement>>();
 
     define(tagName: string, elementClass: CustomElementConstructor) {
       tagName = tagName.toLowerCase();
--- a/packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
+++ b/packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
@@ -29,10 +29,11 @@ interface CustomHTMLElement {
     newValue?: string | null,
     namespace?: string | null
   ): void;
-  formAssociatedCallback?: unknown;
-  formDisabledCallback?: unknown;
-  formResetCallback?: unknown;
-  formStateRestoreCallback?: unknown;
+  formAssociatedCallback?(form: HTMLFormElement | null): void;
+  formDisabledCallback?(disabled: boolean): void;
+  formResetCallback?(): void;
+  // TODO(justinfagnani): what is the type of state? Does it matter?
+  formStateRestoreCallback?(state: unknown, mode: string): void;
 }
 
 interface CustomElementRegistry {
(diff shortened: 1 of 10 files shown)


packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
```diff
@@ -11,6 +11,59 @@
...
+    name: string,
+    oldValue?: string | null,
+    newValue?: string | null,
+    namespace?: string | null
+  ): void;
+  formAssociatedCallback?(form: HTMLFormElement | null): void;
+  formDisabledCallback?(disabled: boolean): void;
+  formResetCallback?(): void;
+  // TODO(justinfagnani): what is the type of state? Does it matter?
+  formStateRestoreCallback?(state: unknown, mode: string): void;
```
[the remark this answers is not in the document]
justinfagnani: thanks!

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
```diff
@@ -391,7 +475,7 @@ if (!ShadowRoot.prototype.createElement) {
       // Approximate observedAttributes from the user class, since the stand-in element had none
       definition.observedAttributes.forEach((attr) => {
         if (instance.hasAttribute(attr)) {
-          definition.attributeChangedCallback.call(
+          definition.attributeChangedCallback!.call(
```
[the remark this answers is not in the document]
justinfagnani: I think the non-null assertion is ok in this case - my main question is why TS thinks this may be null. I want to do another pass to use optional-chaining and rely on null-checks a bit more. I'll come back to it.

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
```diff
@@ -529,27 +634,39 @@ if (!ShadowRoot.prototype.createElement) {
...
-        return this[key];
+      [index: number]: Element;
+
+      ['item'](index: number): Element | null {
+        return this[index];
+      }
+
+      [Symbol.iterator](): IterableIterator<Element> {
+        throw new Error('Method not implemented.');
       }
```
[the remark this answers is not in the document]
justinfagnani: I think that'll grab both named items and number-index items and so duplicate the named items. I think we'll need to keep an array of items to make this work.

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
```diff
@@ -11,6 +11,53 @@
...
+    newValue?: string | null,
+    namespace?: string | null
+  ): void;
+  formAssociatedCallback?: unknown;
+  formDisabledCallback?: unknown;
+  formResetCallback?: unknown;
+  formStateRestoreCallback?: unknown;
+}
+
+interface CustomElementRegistry {
```
rictic, line 41: This is an odd name for this interface, and it doesn't match the global type of the same name

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
```diff
@@ -11,6 +11,53 @@
...
+  disconnectedCallback?: CustomHTMLElement['disconnectedCallback'];
+  adoptedCallback?: CustomHTMLElement['adoptedCallback'];
+  attributeChangedCallback?: CustomHTMLElement['attributeChangedCallback'];
+  formAssociated?: boolean;
+  formAssociatedCallback?: CustomHTMLElement['formAssociatedCallback'];
+  formDisabledCallback?: CustomHTMLElement['formDisabledCallback'];
+  formResetCallback?: CustomHTMLElement['formResetCallback'];
+  formStateRestoreCallback?: CustomHTMLElement['formStateRestoreCallback'];
+  observedAttributes: Set<string>;
+  standInClass?: CustomElementConstructor;
```
rictic, line 57: Add a comment on what standInClass means, and how it's different from elementClass

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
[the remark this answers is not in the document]
bicknellr: My guess is that it can't prove that `definition.attributeChangedCallback` hasn't changed between the first and second property access, since `forEach` and `hasAttribute` might be overridden to do something weird like change the value themselves or delay running the callback until some arbitrary later time. IDK but I doubt TS has special knowledge of / takes into consideration the implementation of built in functions during type checking.

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
```diff
@@ -11,6 +11,61 @@
...
+  ): void;
+}
+
+interface CustomElementRegistry {
+  _getDefinition(tagName: string): CustomElementDefinition;
+}
+
+interface CustomElementDefinition {
+  elementClass: CustomElementConstructor;
+  connectedCallback?: CustomHTMLElement['connectedCallback'];
```
rictic, line 47:
```suggestion
  /**
   * We hold onto the versions of callbacks at registration time, because that's the
   * specc'd behavior.
   */
  connectedCallback?: CustomHTMLElement['connectedCallback'];
```

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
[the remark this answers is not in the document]
bicknellr: I think slice only grabs properties that have number-as-string keys.

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
```diff
@@ -11,6 +11,61 @@
...
+  formStateRestoreCallback?: CustomHTMLElement['formStateRestoreCallback'];
+  observedAttributes: Set<string>;
+  standInClass?: CustomElementConstructor;
+}
+
+interface ShadowRootWithSettableCustomElements extends ShadowRoot {
+  customElements?: CustomElementRegistry;
+}
+
+type ParametersOf<
```
rictic, line 64: How is this different from `Parameters<T>`?

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
```diff
@@ -31,14 +86,12 @@ if (!ShadowRoot.prototype.createElement) {
   // register stand-in elements that can delegate out to CE classes registered
   // in scoped registries
   window.CustomElementRegistry = class {
-    constructor() {
-      this._definitionsByTag = new Map();
-      this._definitionsByClass = new Map();
-      this._whenDefinedPromises = new Map();
-      this._awaitingUpgrade = new Map();
-    }
+    private _definitionsByTag = new Map();
```
rictic, line 89: readonly?

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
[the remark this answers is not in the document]
rictic: It's because we're using forEach instead of `for of`, TypeScript doesn't know when this callback will be called, and whether `definition` will be mutated in the meanwhile

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
```diff
@@ -592,17 +593,19 @@ if (!ShadowRoot.prototype.createElement) {
     // Emulate the native RadioNodeList object
     const RadioNodeList = (class
       extends Array<Node>
-      implements Omit<RadioNodeList, 'forEach' | 'item'> {
+      implements Omit<RadioNodeList, 'forEach'> {
```
rictic, line 598: why not implement 'forEach' as well?

rictic (approved):
A few questions, a few requests for additional comments, otherwise LGTM

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
justinfagnani: I couldn't get the types right :/ The implementation is already there in Array though.

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
justinfagnani: oh... this is a subtle bit that might require a comment: because this file has no imports or exports, it's global.

Here's TypeScript's (not exactly accurate wrt to EcmaScript) docs:

> In TypeScript, just as in ECMAScript 2015, any file containing a top-level import or export is considered a module. Conversely, a file without any top-level import or export declarations is treated as a script whose contents are available in the global scope (and therefore to modules as well).

So this declaration is _adding_ to the existing global, open `CustomElementRegistry` interface.

This is safe against polluting user global types because we don't publish the .d.ts output.

rictic: Yeah, this definitely warrants a big comment at the top of the file

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
```diff
@@ -467,7 +557,7 @@ if (!ShadowRoot.prototype.createElement) {
   ) {
     const internalsToHostMap = new WeakMap();
```
rictic, line 562: Here and elsewhere, every `new Map`, `new WeakMap`, `new Set`, and `new WeakSet` call needs parameters or it's implicitly `any`

rictic (changes requested):

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
rictic: Done.

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
rictic: Done

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
rictic: Ah, it handles undefined

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
rictic: Done

rictic (approved):

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
rictic: Added comment.

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
rictic: Done

PR #572, opened 2024-02-29
Fix closure compiler error

sjamesr:
I'm trying to fix some bugs in the Closure compiler template type checker. The fix has revealed a type error in shadydom, which this PR attempts to fix.

With the fix, the following error occurs:

```
shadydom/src/patch-events.js:286:53: ERROR - [JSC_TYPE_MISMATCH] actual parameter 3 of Object.defineProperty does not match formal parameter
found   : ObjectPropertyDescriptor<Event>
required: ObjectPropertyDescriptor<(Event|{
  composed: ?,
  composedPath: function(this:Event): ?,
  relatedTarget: ?,
  stopImmediatePropagation: function(this:Event): undefined,
  stopPropagation: function(this:Event): undefined,
  target: ?
})>
  286|     /** @type {!ObjectPropertyDescriptor<!Event>} */ ({
                                                            ^^
  287|       /**
       ^^^^^^^^^
...
  296|       configurable: true,
       ^^^^^^^^^^^^^^^^^^^^^^^^^
  297|     })
       ^^^^^^

```

This error is correct because EventPatches is an object literal, which can never be an Event because Event is a nominal type in Closure.

I think this change should be OK and will unblock Closure compiler fixes.

### DIFF
--- a/packages/shadydom/src/patch-events.js
+++ b/packages/shadydom/src/patch-events.js
@@ -283,7 +283,14 @@ if (eventPhaseDescriptor) {
   Object.defineProperty(
     EventPatches,
     'eventPhase',
-    /** @type {!ObjectPropertyDescriptor<!Event>} */ ({
+    /** @type {!ObjectPropertyDescriptor<!{
+  composed: ?,
+  composedPath: function(this:Event): ?,
+  relatedTarget: ?,
+  stopImmediatePropagation: function(this:Event): undefined,
+  stopPropagation: function(this:Event): undefined,
+  target: ?
+}>} */ ({
       /**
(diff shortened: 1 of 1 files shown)


brad4d:
To clarify: In the description the phrase "with the fix the following error occurs" is saying that when closure-compiler is fixed it will report the error unless this PR is accepted first.

Please accept this PR to avoid having this error occur in your build.

brad4d:
I think a better title for this issue would be "Fix a type error that closure-compiler misses now, but will soon start reporting"

rictic (approved):

Issue #580, opened 2024-04-02
[scoped-custom-element-registry] Missing native events on firefox


JGiard:
### Description

Hello,
I was working on a web-component with a <dialog> element, and was having issues with my events that I managed to narrow down to importing scoped-custom-element-registry. The bug only happens on firefox and not on chrome.

I have a web-component built using lit v3 that contains a dialog element. As soon as I add an import to scoped-custom-element-registry, all events dispatched inside the dialog element fail to propagate outside of the web-component

### Example

Here is a minimal html page to reproduce

```<body>
<script type="module">
    import {LitElement, html} from 'https://cdn.jsdelivr.net/gh/lit/dist@3/core/lit-core.min.js';
    import "https://cdn.jsdelivr.net/npm/@webcomponents/scoped-custom-element-registry@0.0.9/src/scoped-custom-element-registry.js"

    export class MyModal extends LitElement {
        render() {
            return html`
                <button @click=${this.open}>open</button>
                <dialog>
                    <slot></slot>
                </dialog>
            `;
        }

        open() {
            this.shadowRoot.querySelector("dialog").showModal()
        }
    }

    customElements.define('my-modal', MyModal);
</script>
<div id="main">
    <my-modal>
        <button>click here</button>
    </my-modal>
</div>
<script>
    document.querySelector("#main").addEventListener("click", (e) => console.log("click on main", e, e.target))
</script>
</body>```

### Steps to reproduce

Click on the button to open the dialog, and click again inside.
Expected behavior is to have two logs "click on main" in the console, one for each click.

On firefox, the second click event does not propagate upwards ouside of the web-component.
Removing the scoped-custom-element-registry import fixes the problem.
This affects all events (not only MouseEvents).

### Version

Lit-element v3
@webcomponents/scoped-custom-element-registry v0.0.9

### Browsers affected

It looks like it only recently stopped working on firefox, a colleague mistakenly had an no-up-to-date firefox v109 and could not reproduce it.

- [ ] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Safari
- [ ] IE 11

BartJanvanAssen:
I'm having the same issue, do you have an update?

BartJanvanAssen:
I was given a workaround via discord, check it out here: https://github.com/webcomponents/polyfills/issues/547#issuecomment-2077323905

Issue #570, opened 2024-01-02
[scoped-custom-element-registry] `importNode is not a function` on latest Chromium


tlouisse:
### Description

The latest version of Chromium (120) supports `ShadowRoot.prototype.createElement`. 
However, it does not support `ShadowRoot.prototype.importNode`.

Since the polyfill uses the existence of `ShadowRoot.prototype.createElement` as a check for polyfill application, we get a console error `importNode is not a function` during lit render (or whenever importNode would have been called).

### Example

 
 Open this in latest Chromium (~120):
 https://4hrpnd-5000.csb.app/index.html

### Steps to reproduce

See  https://codesandbox.io/p/devbox/4hrpnd?file=%2Findex.html%3A11%2C5

1. import the polyfill
2. create a custom element with a scoped registry (we do this with the help of ScopedElementsMixin [in this example]( https://codesandbox.io/p/devbox/4hrpnd?file=%2Findex.html%3A11%2C5) )
3. bootstrap the element in html via its tag name.

#### Expected behavior

The polyfill discriminates between different features that make up the scoped registries spec, (so that the demo works)

#### Actual behavior

Prototype of ShadowRoot is not patched for `importNode`

### Version

0.0.9

### Browsers affected

- [ x ] Chrome
- [ ] Firefox
- [ x ] Edge
- [ ] Safari
- [ ] IE 11

tlouisse:
N.B.: this only occurs when the "Experimental Web Platform features" flag has been set in `chrome://flags`. So it will only affect a smaller portion of people atm (developers like me that forgot this was enabled). However, it would still be good to fix it.

jorenbroekema:
Experienced the same issue. Suddenly, my web app was completely broken, that is, every UI component stopped rendering. Fortunately could fix it by disabling the feature flag but it's very worrying that a portion of my customer-base will experience a broken site. In my opinion fixing this is quite urgent :\

Issue #547, opened 2023-07-24
[scoped-custom-element-registry]: Disabled attribute blocks any click on custom elements (all the WebComponents are form-associated)


igomezal:
### Description

WebComponents defined with the scoped-custom-element-registry polyfill not receive any click event when they have the disabled attribute set.

The polyfill seems to be setting the `formAssociated` static getter to all the WebComponents registered with the polyfill enabled which cause all the WebComponents to behave like an input, button, textarea, etc.

https://github.com/webcomponents/polyfills/blob/90cb97f847ce918289dac0978c50dcda0a0afd72/packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js#L217-L220

As it enables the `formAssociated` property, it blocks any click from the user when the component is disabled something that should only happen on real form components like [button](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button), [fieldset](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset) , [optgroup](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/optgroup) , [option](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/option) , [select](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select), [textarea](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea) and [input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input) as defined in https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled

This is not the only issue detected, other users also seems to find strange behaviours with the polyfill enabled like in the issue https://github.com/webcomponents/polyfills/issues/546

### Example

```html
<!DOCTYPE html>
 <html>
 <head>
   <script src="https://www.unpkg.com/@webcomponents/scoped-custom-element-registry@0.0.9/scoped-custom-element-registry.min.js"></script>
 </head>
 <body>
   <my-custom-webcomponent disabled></my-custom-webcomponent>

   <script>
     class MyCustomWebComponent extends HTMLElement {
       constructor() {
         super()

         const shadow = this.attachShadow({ mode: 'open' })

         shadow.innerHTML = '<p>Click here!</p>'
       }

       connectedCallback() {
        this.addEventListener('click', this.alertOnClick);
       }

       disconnectedCallback() {
        this.removeEventListener('click', this.alertOnClick);
       }

       alertOnClick() {
        alert('YOU CLICKED!');
[...]

PR #581, opened 2024-05-06
[scoped-custom-element-registry] Improve issues related to `formAssociated`

### LINKED ISSUE #546: [scoped-custom-element-registry]: Safari v16.4 focuses all custom elements
### Description

Safari v16.4 supports `attachInternals` and `formAssociated` features natively, and with the subject polyfill enabled _all_ custom elements on the page are keyboard-focusable using Tab key, regardless of the `role` or anything else.

Here is the root of the problem:

https://github.com/webcomponents/polyfills/blob/90cb97f847ce918289dac0978c50dcda0a0afd72/packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js#L218-L220

Feels like Safari tries to focus any element with `static formAssociated = true`. If I change that to `false` in the polyfill source code then the above wrong behavior becomes fixed.

Is there a way to respect the original `formAssociated` value instead of enforcing it to be `true`?

### Example

```html
<!DOCTYPE html>
 <html>
 <head>
   <script src="https://www.unpkg.com/@webcomponents/scoped-custom-element-registry@0.0.9/scoped-custom-element-registry.min.js"></script>
 </head>
 <body>
   <input type="text"/>
   <my-el></my-el>
 
   <script>
     class MyEl extends HTMLElement {
       constructor() {
         super()
 
         const shadow = this.attachShadow({ mode: 'open' })
 
         shadow.innerHTML = '<span>Hello</span>'
       }
     }
 
     customElements.define('my-el', MyEl)
   </script>
 </body>
 </html>
```

### Version

v0.0.9

### Browsers affected

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [x] Safari
- [ ] IE 11

sorvell:
Fixes #546, #547, #580, #570, #549. 

Note, bulk of changes in [this commit](https://github.com/webcomponents/polyfills/pull/581/commits/b0ef750fe9e4759a3c6202d05ae5c2c7e9ebe203).

The setting for a defined class' `formAssociated` cannot be properly polyfilled per registry. To address this, previously the polyfill always set the stand-in class to be `formAssociated == true`. However, this caused a series of subtle issues.

Instead, it is now set to whatever the first defined tag name's `formAssociated` is set. It's also possible to reserve a given name to always be formAssociated by adding it to `customElementRegistryPolyfill.formAssociated.add(tagName)`.

Unconditionally installs polyfill.

Adds `registry` to ShadowRootInit


packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js
```diff
@@ -11,12 +11,22 @@
  * subject to an additional IP rights grant found at
  * http://polymer.github.io/PATENTS.txt
  */
-if (!ShadowRoot.prototype.createElement) {
+if (!(ShadowRoot.prototype.createElement && ShadowRoot.prototype.importNode)) {
```
thepassle, line 14: Should this feature check be here at all?

sorvell: Reverted for now. Will have more to change/discuss to potentially better support Chrome's partial implementation in a follow-on.

sorvell: Based on dicussion with @justinfagnani and suggesting in #549, check was removed.

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
```diff
@@ -19,6 +19,22 @@
...
  */
 
+interface Window {
+  CustomElementRegistryPolyfill: {
+    formAssociated: Set<string>;
+  };
+}
+
+// Polyfill helper object
+(window as Window)['CustomElementRegistryPolyfill'] = {
```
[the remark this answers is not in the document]
sorvell: Done and added comment.

Issue #609, opened 2025-02-25
[scoped-custom-element-registry] Broken on Safari 18.4


MariusVatasoiu:
### Description

Error thrown on Safari 18.4  at this [line](https://github.com/webcomponents/polyfills/blob/90cb97f847ce918289dac0978c50dcda0a0afd72/packages/scoped-custom-element-registry/src/scoped-custom-element-registry.js#L411C5-L411C66):

```js
const shadowRoot = nativeAttachShadow.apply(this, arguments);
```

### Steps to reproduce

```js
import "https://esm.sh/@webcomponents/scoped-custom-element-registry@0.0.9"
import { LitElement, html } from "https://esm.sh/lit"
import { ScopedElementsMixin } from "https://esm.sh/@open-wc/scoped-elements/html-element.js"

class Child extends LitElement {
  render() {
    return html`child`
  }
}

class MyComponent extends ScopedElementsMixin(LitElement) {
  static scopedElements = {
    "my-child": Child,
  }
  render() {
    return html`<my-child></my-child>`
  }
}

window.customElements.define("my-component", MyComponent)
```

```html
<my-component></my-component>
```

Repro: [playground](https://jsfiddle.net/MariusVatasoiu/ahe4gkfs/)

#### Expected behavior

No error is thrown

#### Actual behavior

Error is thrown

### Version

@webcomponents/scoped-custom-element-registry@0.0.9

### Browsers affected

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [x] Safari 18.4 (beta)
- [ ] IE 11

MariusVatasoiu:
I've tested the code from master branch and the problem seems to be solved. If I'm not wrong, this https://github.com/webcomponents/polyfills/pull/581 PR from @sorvell solves the issue, so technically we just need a new release.

justinfagnani:
@rniwa No one's filed a WebKit bug yet, but I wanted to tag you here as a heads up.

annevk:
Does someone have a more minimal test case? As far as I know Scoped Custom Element Registries is not enabled in 18.4 so I wonder why that would be the cause.

MariusVatasoiu:
I left a repro in the first post:

https://jsfiddle.net/MariusVatasoiu/ahe4gkfs/

annevk:
I saw that, I meant without all the library code.

sorvell:
[...]

PR #611, opened 2025-02-26
[scoped-custom-element-registry] Fix TypeError in WebKit

justinfagnani:
WebKit's prototype native scoped elements registry doesn't like getting a `registry` key in `attachShadow()`.

The previous state did attempt to remove `registry`, but the key was getting renamed by Closure.

Fixes https://github.com/webcomponents/polyfills/issues/609

This PR also updates Playwright and enables WebKit Playwright tests in the scoped-custom-element-registry package. Other repo tests are failing on CI, so CI isn't any help here. The scoped-custom-element-registry package tests are passing locally for me.

edit: I should have noted that WebKit had 23 test failures before the fix. All passing after the fix.

### DIFF
--- a/packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
+++ b/packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
@@ -666,22 +666,22 @@ const customize = (
 const nativeAttachShadow = Element.prototype.attachShadow;
 Element.prototype.attachShadow = function (
   init: ShadowRootInitWithSettableCustomElements,
-  ...args: unknown[]
+  ...args: Array<unknown>
 ) {
-  // note, set registry to undefined to avoid using a native implementation
-  const nativeArgs = ([
-    {
-      ...init,
-      registry: undefined,
-    },
-    ...args,
-  ] as unknown) as [init: ShadowRootInit];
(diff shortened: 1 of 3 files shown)


packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
```diff
@@ -666,22 +666,22 @@ const customize = (
...
-    ...args,
-  ] as unknown) as [init: ShadowRootInit];
-  const shadowRoot = nativeAttachShadow.apply(this, nativeArgs);
-  const registry = init['registry'] ?? init.customElements;
+  // Note, We must remove `registry` from the init object to avoid passing it to
+  // the native implementation. Use string keys to avoid renaming in Closure.
+  const {
+    'customElements': customElements,
+    'registry': registry = customElements,
+    ...nativeInit
```
sorvell, line 676: What's this?

justinfagnani: rest object for destructuring assignment. It declares `newInit` which gets every property in `init` _except` `customElements` and `registry`.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#rest_property

PR #625, opened 2025-07-17
fix: added a workaround for NordVPN browser protection

### LINKED ISSUE #624: [scoped-custom-elements-poilyfill] NordVPN browser extension breaks properties registration on ShadowRoot
### Description

When using the `scoped-custom-element-registry` polyfill in combination with the NordVPN browser extension, the extension rewrites or tampers with the ShadowRoot prototype chain. As a result, code that relies on ShadowRoot.host to access the scoped registry of the custom element fails.

### Example

It can be reproduced on https://ing.be

### Steps to reproduce

1. To the the website
2. Inspect in the Elements experience-renderer
3. Inspect its ShadowRoot
4. Check in the console registry property

#### Expected behavior

No NordVPN extension:
I {h: Map(5), m: Map(5), j: Map(0), i: Map(0)}

#### Actual behavior

NordVPN extension:
undefined

### Version

0.0.10

### Browsers affected
Only tested in Chrome

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari
- [ ] IE 11

paulcodiny:
Added a workaround to look up the local "registry" on a host element apart from the Shadow Root

Fixes #624

### DIFF
--- a/packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
+++ b/packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
@@ -372,8 +372,11 @@ const registryForNode = (node: Node): ShimmedCustomElementsRegistry | null => {
       scope = scopeForElement.get(scope)?.getRootNode() || document;
     }
   }
+  // @see https://github.com/webcomponents/polyfills/issues/624
+  // The problem is that NordVPN add "registry" property on the host element
+  // this can break if the host element has a "registry" property already but there is not solution for this
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  return (scope as any)['registry'] as ShimmedCustomElementsRegistry | null;
+  return ((scope as any)['registry'] || ((scope as any).host ? (scope as any).host['registry'] : null)) as ShimmedCustomElementsRegistry | null;
 };
 
 // Helper to create stand-in element for each tagName registered that delegates


packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
```diff
@@ -372,8 +372,11 @@ const registryForNode = (node: Node): ShimmedCustomElementsRegistry | null => {
       scope = scopeForElement.get(scope)?.getRootNode() || document;
     }
   }
+  // @see https://github.com/webcomponents/polyfills/issues/624
+  // The problem is that NordVPN add "registry" property on the host element
+  // this can break if the host element has a "registry" property already but there is not solution for this
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  return (scope as any)['registry'] as ShimmedCustomElementsRegistry | null;
+  return ((scope as any)['registry'] || ((scope as any).host ? (scope as any).host['registry'] : null)) as ShimmedCustomElementsRegistry | null;
```
justinfagnani, line 379: Can you describe the problem and this solution a little more?

If NordVPN adds `registry` to the element (is that every element? Element.prototype? Some elements?), why would it interfere with `scope.registry`? Is it because NordVPN adds a particular kind of value to `registry`? What is it?

`scope` is usually a Document or ShadowRoot here. Is this only a problem when using innerHTML into a disconnected tree so that `scope` is an element? If so, could this be better handled in the if case above (line 372) where we know we have an Element and not a Document or ShadowRoot.

It's not obviously ok to me to be falling back to a ShadowRoot's host element to look for a registry, so I'm a little nervous about this change and want to understand it better.

justinfagnani (changes requested):

packages/scoped-custom-element-registry/src/scoped-custom-element-registry.ts
sorvell: Yeah, I also don't understand either the problem or solution.
* #624 says the extension tampers with ShadowRoot's prototype but it doesn't say how exactly.
* the code here is falling back to looking for a registry on the host. That's not something the polyfill sets so where is that coming from? The extension?

We also need a test that sets up the DOM similar to the extension and demonstrates that this fix makes something work that otherwise wouldn't.

sorvell (changes requested):
Please add a test setting up the DOM in the same way the extension does and demonstrate that the change here fixes it.

Please also change the comment in the code to explain exactly how the extension breaks an expectation and how the code here fixes it.

